PR #193, opened 2025-09-15
Users/beaub/dotnet provider args

beaubelgrave:
This PR enables passing filter arguments to providers within DotNet on both Windows and Linux, however, only the recording/consumption side of this is fully enabled on Linux, since that is a higher priority.

Later PR will enable the Windows side, as well as parsing the dynamic event data formats. Right now, when provider levels are recorded, the events only contain a default record type. Fields are not known or recorded into the file formats. This means consumers must know the format details (Fine for manifest events, not great for self-describing events).

However, this gets us very close to our goal of being able to record a wide set of events on several providers with only scripts having to enter in the provider name and the keyword + level details, instead of every unique event. This also enables scripts to be able to collect DotNet counters via the System.Runtime provider, which requires filter arguments.

### DIFF
--- a/one_collect/src/helpers/dotnet/os/linux.rs
+++ b/one_collect/src/helpers/dotnet/os/linux.rs
@@ -11,6 +11,7 @@ use std::io::{Read, BufRead, BufReader, Write};
 use std::fs::File;
 use std::path::{Path, PathBuf};
 use std::collections::{HashSet, HashMap};
+use std::collections::hash_map::Entry::{Occupied, Vacant};
 use std::sync::mpsc::{self, Sender, Receiver};
 use std::sync::{Arc, Mutex};
 use std::thread::{self, JoinHandle};
@@ -18,7 +19,10 @@ use std::thread::{self, JoinHandle};
 use crate::helpers::dotnet::*;
 use crate::helpers::dotnet::universal::UniversalDotNetHelperOSHooks;
 use crate::helpers::exporting::{UniversalExporter, ExportSettings, ExportTraceContext};
+use crate::helpers::exporting::process::MetricValue;
+use crate::helpers::exporting::record::ExportRecordType;
 
+use crate::intern::InternedStrings;
 use crate::user_events::*;
 use crate::tracefs::*;
 use crate::perf_event::*;
@@ -177,32 +181,15 @@ impl UserEventProviderEvents {
         count
     }
 
-    fn ensure_keyword_level(
-        &mut self,
-        keyword: u64,
-        level: u8) {
-        self.keyword |= keyword;
-
-        if level > self.level {
-            self.level = level;
-        }
-    }
-
     fn add(
         &mut self,
         tracepoint: String,
-        dotnet_events: &HashSet<usize>,
-        keyword: u64,
-        level: u8) {
+        dotnet_events: &HashSet<usize>) {
         let mut events = Vec::new();
 
         for event in dotnet_events {
             events.push(*event as u32);
         }
-
-        self.ensure_keyword_level(
-            keyword,
-            level);
 
         self.events.push(
             UserEventTracepointEvents {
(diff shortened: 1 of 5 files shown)


one_collect/src/helpers/dotnet/os/linux.rs
```diff
@@ -840,29 +1112,42 @@ impl OSDotNetEventFactory {
...
-                        &dotnet_ids,
-                        keyword,
-                        level);
+                        &dotnet_ids);
+                }
+
+                if provider.borrow().has_named_events() ||
+                   provider.borrow().has_callbacks() {
+                    let tracepoint = format!(
+                        "OC_DotNet_{}_{}_EventSource",
```
brianrob, line 1137: I think I missed this when reviewing.  I see that this is where we're creating tracepoints to flow EventSource events - what makes this path EventSource specific?

one_collect/src/helpers/dotnet/scripting/mod.rs
```diff
@@ -159,6 +159,38 @@ impl DotNetEventGroup {
     }
 }
 
+#[derive(Default, Clone)]
+pub (crate) struct DotNetProviderFlags {
+    callstacks: bool,
+    callstack_keywords: u64,
```
brianrob, line 165: Is `callstack_keywords` enabling the .NET runtime stack event, or is this filtering callstacks to a particular keyword?

brianrob (commented):
In addition to the questions below - I have a couple of overarching questions:

- Is the goal of `record_dotnet_provider` to enable us to keep track of the providers that need filter args and direct them appropriately, or is there more to it than that?
- I know that we need to be able to handle self-describing events for `EventSource` - is `self_describing_event_from_dotnet` needed to enable the event and proxy it to a set of callbacks, or is this for something else?

one_collect/src/etw/mod.rs
```diff
@@ -1074,7 +1081,7 @@ impl EtwSession {
             }
         }));
 
-        let context = SessionCallbackContext::new(0);
+        let context = SessionCallbackContext::new(0, 0);
```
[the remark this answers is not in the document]
beaubelgrave: Keeping the session ID would be fine, however, the handle at that point is invalid since the session is stopped. If folks use the session ID for some lookup, I guess it might be nice to have it upon stopped. The DotNet helper does this, but it clears all the data at the stopping point instead of stopped point. I think it's most safe to keep the session ID around during stopped incase folks cannot use stopping for their needs.

one_collect/src/helpers/dotnet/os/linux.rs
beaubelgrave: I was under the impression that named events are only from EventSource, since the manifest events do not give back metadata (and therefore won't have a name). The EventSource part is due to the has_named_events() being true, which means that scripts registered a self describing event.

If there are self describing events that are outside of EventSource, I'm happy to change the name of the tracepoint to just NamedEvents or even just All since this tracepoint is the default tracepoint for the IPC session. It does not get used for events that have IDs, only for events with Names or if provider wide callbacks are enabled (IE: record_dotnet_provider).

one_collect/src/helpers/dotnet/scripting/mod.rs
beaubelgrave: Just filtering callstacks to a particular keyword, this is the same mechanism as ETW on Windows, so I feel safe in allowing this at the DotNet level, since we can support this on both OS's equally.

one_collect/src/etw/mod.rs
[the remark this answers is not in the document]
brianrob: Sounds good.  Agreed that the handle is invalid - I just imagine people using it as a durable id for something.

one_collect/src/helpers/dotnet/os/linux.rs
brianrob: Sounds good.  Otherwise, good to go.  I'll watch for the commit and approve.

PR #195, opened 2025-09-18
Include underlying error in tracefs error message

### LINKED ISSUE #189: Expose underlying tracefs error in error message
I tried to record a trace with .NET events with

```
record-trace --on-cpu --script-file dotnet.script --out out.nettrace
```

and got

```
Error: TraceFS is not accessible.
```

I found the underlying error by checking the source. Including it in the error message would've made it actionable for me, e.g.,

```
Error: Tracefs is not accessible: It appears tracefs is not mounted. You can mount it by running mount -t tracefs nodev /sys/kernel/tracing.
```

I have a branch ([diff](https://github.com/microsoft/one-collect/compare/main...zachcmadsen:one-collect:tracefs-error?expand=1)) with a change for this if you think it's reasonable.

beaubelgrave:
closes #189

### DIFF
--- a/one_collect/src/helpers/dotnet/os/linux.rs
+++ b/one_collect/src/helpers/dotnet/os/linux.rs
@@ -728,19 +728,16 @@ impl OSDotNetEventFactory {
         &mut self,
         exporter: UniversalExporter) -> UniversalExporter {
         let fn_providers = self.providers.clone();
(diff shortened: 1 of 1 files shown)


brianrob (approved):
LGTM. Thanks!

PR #196, opened 2025-09-18
Security: Replace rust-crypto with sha1 crate

beaubelgrave:
The rust-crypto crate is no longer maintained and uses an old version of the time crate that can segfault under strict conditions (including supply chain attacks).

Remove rust-crypto and pull in sha1 that is actively maintained and does not include time or other unneeded crates.

Update dotnet scripting use of Sha1 for EventSource GUID calculation.

Closes https://github.com/microsoft/one-collect/security/dependabot/15 (and others).


brianrob (commented):
LGTM.  Please also remove rust-crypto from https://github.com/microsoft/one-collect/blob/main/THIRD-PARTY-NOTICES.md#rust-crypto and add sha1.

brianrob (approved):

PR #197, opened 2025-09-18
record-trace: Add start/end to output types

beaubelgrave:
Previously record-trace would assume CTRL+C is the stopping mechanism. Now that we have an FFI, we cannot assume that detail. We need a way for solutions based on the FFI to communicate how to cancel (or any other details like distro information) when the recording starts.

Add with_start() and with_end() to EngineOutput to allow Rust tools to hook start/end.

Hook with_start() and with_end() to FFI using new values (OUTPUT_START/END) passed to the callback.

Remove CTRL+C from engine crate and move that banner printing to record-trace CLI to ensure we do not confuse folks. Each tool needs to communicate how to cancel the recording (if at all).


brianrob (approved):
LGTM.  Thanks!

PR #198, opened 2025-09-19
Fix ELF Symbol Resolution

brianrob:
Fixes three issues found in ELF symbol resolution, which were causing incorrect symbols to be resolved for some binaries.

- The PT_LOAD header was being fetched from whichever file contained the symbols, but the PT_LOAD header values are only valid in the actual binary and not in sidecar symbol files.
- The PT_LOAD header values need to be page aligned when computing the file offsets of the symbol start and end values.
- Removed special handling for SHT_NOBITS (.bss) segments because it was breaking symbol resolution and is not used for function symbols.

### DIFF
--- a/one_collect/src/helpers/exporting/modulemetadata.rs
+++ b/one_collect/src/helpers/exporting/modulemetadata.rs
@@ -48,6 +48,8 @@ pub struct ElfModuleMetadata {
     build_id: Option<[u8; 20]>,
     debug_link_id: usize,
     version_metadata_id: usize,
+    p_vaddr: u64,
+    p_offset: u64,
 }
 
 impl ElfModuleMetadata {
@@ -56,6 +58,8 @@ impl ElfModuleMetadata {
             build_id: None,
             debug_link_id: 0,
             version_metadata_id: 0,
+            p_vaddr: 0,
+            p_offset: 0,
         }
     }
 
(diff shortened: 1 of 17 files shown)


one_collect/src/helpers/exporting/os/linux.rs
```diff
@@ -34,6 +34,8 @@ pub type Session = PerfSession;
 /* OS Specific Session Builder Type */
 pub type SessionBuilder = RingBufSessionBuilder;
 
+const PAGE_MASK: u64 = 0xFFFFFFFFFFFFF000;
```
beaubelgrave, line 37: Can we assume 4K pages here or do we need to load this from the binary or operating system?

one_collect/src/helpers/exporting/modulemetadata.rs
```diff
@@ -48,6 +48,8 @@ pub struct ElfModuleMetadata {
     build_id: Option<[u8; 20]>,
     debug_link_id: usize,
     version_metadata_id: usize,
+    p_vaddr: u64,
+    p_offset: u64,
```
beaubelgrave, line 52: Do we need to output these in NetTrace (and other formats) for offline symbol resolution? Or is offline resolution unaffected by this new calculation?

brianrob: Yes.  I've added these to the version metadata.

PR #200, opened 2025-09-24
PageFaults: Add soft/hard page faults support

### LINKED ISSUE #186: Add perf_event major/minor page faults as events
Perf_event ABI allows to capture major and minor page faults. The existing tracepoint exceptions/page_fault_user includes both, but often times we only want major faults (especially when stacks are being collected). We need to allow for this like we do cswitch events.

beaubelgrave:
Users often want to know if soft (non-IO) or hard (IO) page faults are occurring. Both Windows and Linux support reporting this data.

Add enabling soft/hard page faults to ExportSettings.

Add soft/hard page fault event to ExportMachine.

Have each OS trigger the soft/hard page fault event.

Add soft/hard page fault settings within record-trace.

Closes #186


one_collect/src/helpers/exporting/os/linux.rs
```diff
@@ -692,11 +692,58 @@ impl OSExportMachine {
...
+        let ancillary = session.ancillary_data();
+        let time_field = session.time_data_ref();
+        let pid_field = session.pid_field_ref();
+        let tid_field = session.tid_data_ref();
+        let reader = callstack_reader.clone();
+
+        /* Get sample kind for event */
+        let kind = machine.borrow_mut().sample_kind(kind);
+
+        /* Hook hard page fault event */
```
brianrob, line 709: I think this comment is a copy-paste bug since this is now in a separate function.

brianrob (commented):
LGTM.  This is a great change!  One small note below - not blocking.

PR #201, opened 2025-09-25
Specify the Application Name in the Parser

brianrob:
This is required now that record-trace is split into multiple crates. Without this, `record-trace -v` shows the name of the application as "engine".


beaubelgrave (approved):
LGTM, thanks!

PR #207, opened 2025-10-15
DotNet: Ensure right format ID for dynamic events

beaubelgrave:
We need to differentiate between events that are named and not named when caching the format ID on Linux. When non-dynamic events come in, the metadata may be present, but if there is no name associated with it, we must treat it as a manifest event.

Update format_key to use low 31 bits for the ID, and the top bit to indicate named or not named when doing lookups/caching of format details (including the event ID).

Do not set name_id to Some when the event name is empty.

### DIFF
--- a/one_collect/src/helpers/dotnet/os/linux.rs
+++ b/one_collect/src/helpers/dotnet/os/linux.rs
@@ -575,11 +575,16 @@ struct LinuxDotNetProviderContext<'a> {
 }
 
 impl<'a> LinuxDotNetProviderContext<'a> {
-    fn format_key(&self) -> usize {
-        let name_id = self.info.name_id.unwrap_or(0);
-        let format_id = self.info.format_index.unwrap_or(0);
+    fn format_key(&self) -> u64 {
+        /*
+         * If we have a name use that, otherwise use the
+         * actual event ID with the top bit set so we
+         * don't collide with name IDs.
+         */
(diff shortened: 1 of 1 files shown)


one_collect/src/helpers/dotnet/os/linux.rs
```diff
@@ -576,10 +576,15 @@ struct LinuxDotNetProviderContext<'a> {
...
+        /*
+         * If we have a name use that, otherwise use the
+         * actual event ID with the top bit set (so we
+         * don't collide with name IDs.
+         */
+        let id = self.info.name_id.unwrap_or(1 << 31 | self.id);
         let format_id = self.info.format_index.unwrap_or(0);
 
-        name_id << 32 | format_id
+        id << 32 | format_id
```
brianrob: This returns `usize` won't the top bit overflow here on 32-bit architectures?

beaubelgrave: I should change to u64 explicitly, but we don't support 32-bit architectures in many places.

beaubelgrave: I'll fix it.

brianrob: That's a good point - we can follow-up on a discussion of this.

brianrob (approved):

PR #209, opened 2025-10-24
Linux Target pids should work with existing and future tasks within processes

### LINKED ISSUE #192: [.NET][Record-Trace] --pid not working with .NET Threads
No .NET events are being captured when specifying the .NET process' PID with `--pid` if the .NET process spawns managed threads. Without specifying the `--pid`, events are captured.

```C#
using System.Threading;

for (int i = 0; i < 5; i++)
{
    Thread thread = new Thread(() =>
    {
        while (true) {
            try {
                throw new Exception("Test");
            }
            catch {}
        }
    });
    thread.Start();
}
```

```rust
let scenario = new_dotnet_scenario();
scenario.with_exceptions();
use_dotnet_scenario(scenario);
```

```sh
mihw@CPC-mihw-6KMZDM:~/repo/one-collect/record-trace/target/release$ sudo ./record-trace --pid 90176 --script-file ~/Use
rEventsTest/pidtest.script --out pid-test-stress-only.nettrace --live
Recording started.  Press CTRL+C to stop.
```

```
mihw@CPC-mihw-6KMZDM:~/repo/one-collect/record-trace/target/release$ sudo ./record-trace --script-file ~/Use
rEventsTest/pidtest.script --out pid-test-stress-only.nettrace --live
+1.0409: Microsoft-Windows-DotNETRuntime:{E13C0D23-CCBC-4E12-931B-D9CC2EEE27E4}/ExceptionThrown(cpu_gauge, PID=90176): 1 Count
+1.0409: Microsoft-Windows-DotNETRuntime:{E13C0D23-CCBC-4E12-931B-D9CC2EEE27E4}/ExceptionThrown(cpu_gauge, PID=90176): 1 Count
+1.0410: Microsoft-Windows-DotNETRuntime:{E13C0D23-CCBC-4E12-931B-D9CC2EEE27E4}/ExceptionThrown(cpu_gauge, PID=90176): 1 Count
+1.0410: Microsoft-Windows-DotNETRuntime:{E13C0D23-CCBC-4E12-931B-D9CC2EEE27E4}/ExceptionThrown(cpu_gauge, PID=90176): 1 Count
```

beaubelgrave:
Closes #192.

Perf_event_open() requires an FD per-PID + CPU + Tracepoint, which causes a lot of FDs, so this change also includes setting the rlimit default limit to the max to ensure we can have as many files open as possible.

I've tested scenarios where existing threads are already running and we get data back for them correctly.

I've also tested a process without any extra threads getting collection started, and then having new threads spawn.

In both scenarios I see the new threads data showing up via nettrace within PerfView.

### DIFF
--- a/one_collect/src/perf_event/rb/source.rs
+++ b/one_collect/src/perf_event/rb/source.rs
@@ -463,6 +463,24 @@ impl RingBufDataSource {
         Ok(())
     }
 
+    fn tasks_for_pids(pids: &mut Vec<i32>) {
+        let mut tasks = HashSet::new();
+
+        /* Find all unique tasks IDs */
+        for pid in pids.drain(..) {
+            tasks.insert(pid);
+
+            procfs::iter_proc_tasks(
+                pid as u32,
+                |task| { tasks.insert(task as i32); });
+        }
+
+        /* Update PIDs with unique tasks */
+        for task in tasks.drain() {
+            pids.push(task);
+        }
+    }
+
     fn build(&mut self) -> IOResult<()> {
         /* Always required */
         let common = self.kernel_builder
@@ -471,8 +489,15 @@ impl RingBufDataSource {
 
         let empty_pids = Vec::new();
 
-        let pids = match &self.target_pids {
-            Some(pids) => { pids },
+        let target_pids = &mut self.target_pids.as_mut();
+
+        let pids = match target_pids {
+            Some(pids) => {
+                /* Populate current tasks for PIDs */
+                Self::tasks_for_pids(pids);
+
+                pids
+            },
             None => { &empty_pids },
         };
 
--- a/one_collect/src/perf_event/mod.rs
+++ b/one_collect/src/perf_event/mod.rs
@@ -269,6 +269,18 @@ impl Drop for PerfSession {
 impl PerfSession {
     pub fn new(
         source: Box<dyn PerfDataSource>) -> Self {
+        /* Increase rlimit for open files */
+        unsafe {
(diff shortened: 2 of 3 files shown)


one_collect/src/perf_event/rb/source.rs
```diff
@@ -463,6 +463,24 @@ impl RingBufDataSource {
         Ok(())
     }
 
+    fn tasks_for_pids(pids: &mut Vec<i32>) {
+        let mut tasks = HashSet::new();
+
+        /* Find all unique tasks IDs */
+        for pid in pids.drain(..) {
+            tasks.insert(pid);
```
brianrob: If we insert `pid` here, then we're going to end up with duplicate pids when this function returns, no?

one_collect/src/perf_event/rb/source.rs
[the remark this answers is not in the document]
beaubelgrave: Yep, inherit will do the redirection within the kernel for fork/clone. We unfortunately have to keep the FD in memory for perf_events to keep the redirect event active (I wish this wasn't the case). So we end up with a lot of FDs. Ideally, we'd only need 1 FD per-CPU and have perf_event ref-count the redirected ring buffers, but that's not how it currently works. I'll be poking around with tracing kernel folks if we can make this better over time.

one_collect/src/perf_event/rb/source.rs
brianrob: You're right - I missed that it calls `drain` to empty them out.

PR #229, opened 2026-01-27
Detect DotNet processes with IPC mmap operation

### LINKED ISSUE #226: [record-trace][.NET] Discover .NET apps diagnostics ports without perfmaps
Record-Trace leverages doublemapper/perfmap to discover .NET process PIDs and their diagnostic-ports. https://github.com/microsoft/one-collect/blob/0f08373998ef4b12e3fbc75b46dbdf3a41d52207/one_collect/src/helpers/dotnet/os/linux.rs#L386-L387

Only CoreCLR uses doublemapping, so nativeAOT apps never receive an IPC command from record-trace to emit userevents.

### Repro
record-trace script
```rs
let Microsoft_Windows_DotNETRuntime_flags = new_dotnet_provider_flags();
record_dotnet_provider("Microsoft-Windows-DotNETRuntime", 0x80000000000, 4, Microsoft_Windows_DotNETRuntime_flags);
```
.NET csproj
```csproj
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>net10.0</TargetFramework>
    <ImplicitUsings>enable</ImplicitUsings>
    <Nullable>enable</Nullable>
    <PublishAot>true</PublishAot>
    <SelfContained>true</SelfContained>
    <InvariantGlobalization>true</InvariantGlobalization>
    <StripSymbols>true</StripSymbols>
    <EnableTrimAnalyzer>true</EnableTrimAnalyzer>
    <EventSourceSupport>true</EventSourceSupport>
  </PropertyGroup>

</Project>
```
.NET App
```C#
using System.Diagnostics;

byte[] s_array;

while (true)
{
	s_array = new byte[1024 * 100];
	Thread.Sleep(100);
}
```
`dotnet publish -c Release -r linux-x64`

`./bin/Release/net10.0/linux-x64/publish/<app>`

In a separate terminal window trace for a few seconds
`sudo record-trace --script-file <rust script above>`

### Expected

The generated .nettrace contains `Microsoft-Windows-DotNETRuntime/Unknown(303)` from the .NET app

### Actual

The event is missing, and the process even is listed in ExistedProcess/PerfView's Process view.

brianrob:
One-collect currently depends upon the existence of an mmap call in the .NET runtime for a mapping called doublemapper in order to detect that the process is a .NET process.  This is fragile as doublemapper is an implementation detail used by the W^X implementation.

Rather than depending on this mapping, depend upon the existence of a mapping called dotnet_ipc_created, which will be created by the .NET runtime once the IPC channel has been created.

For the purposes of backwards compatibility, we'll keep checking for doublemapper, though ideally the new mapping is backported.

Fixes #226

### DIFF
--- a/one_collect/src/helpers/dotnet/os/linux.rs
+++ b/one_collect/src/helpers/dotnet/os/linux.rs
@@ -1418,11 +1418,23 @@ impl DotNetHelp for RingBufSessionBuilder {
         };
 
         self.with_hooks(
-            move |_builder| {
-                /* Nothing to build */
+            move |builder| {
+                /* Enable all mmap records (including non-executable) for dotnet helper
+                 * to capture dotnet processes that emit the /memfd:dotnet_ipc_created mapping. */
+                let kernel = builder
+                    .take_kernel_events()
+                    .unwrap_or_else(RingBufBuilder::for_kernel)
+                    .with_all_mmap_records();
+
+                builder.replace_kernel_events(kernel);
             },
 
             move |session| {
+                /* Enable capture of all mmap records (including non-executable) during environment
+                 * capture to discover dotnet processes that have created the /memfd:dotnet_ipc_created mapping. */
+                session.capture_env_options_mut()
+                    .with_all_mmaps();
+
                 if perf_maps {
                     /* Perf map support */
                     let event = session.mmap_event();
--- a/one_collect/src/helpers/callstack/os/linux.rs
+++ b/one_collect/src/helpers/callstack/os/linux.rs
@@ -533,7 +533,7 @@ impl CallstackHelp for RingBufSessionBuilder {
                 let events = builder
                     .take_kernel_events()
                     .unwrap_or_else(RingBufBuilder::for_kernel)
-                    .with_mmap_records()
+                    .with_executable_mmap_records()
                     .with_comm_records()
                     .with_task_records();
 
--- a/one_collect/src/helpers/exporting/os/linux.rs
+++ b/one_collect/src/helpers/exporting/os/linux.rs
@@ -1528,7 +1528,7 @@ impl ExportBuilderHelp for RingBufSessionBuilder {
         let mut builder = self;
 
         let mut kernel = RingBufBuilder::for_kernel()
-            .with_mmap_records()
+            .with_executable_mmap_records()
             .with_comm_records()
             .with_task_records();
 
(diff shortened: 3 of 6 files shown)


beaubelgrave (approved):

one_collect/src/helpers/dotnet/os/linux.rs
mdh1418: Could record-trace also look for non-executable mappings to find `dotnet_ipc_created`? https://github.com/dotnet/runtime/pull/123779/changes#r2743888505

one_collect/src/helpers/dotnet/os/linux.rs
```diff
@@ -1297,7 +1297,7 @@ impl OSDotNetEventFactory {
                         let filename = fmt.get_str(filename, data)?;
 
                         /* Check if dotnet process */
-                        if filename.starts_with("/memfd:doublemapper") {
+                        if filename.starts_with("/memfd:doublemapper") || filename.starts_with("/memfd:dotnet_ipc_created") {
```
jkotas: How much of backward compatibility do we care about here? What are the .NET runtime versions that we expect this to work on?

If we really need to keep this around, it should have a comment that the doublemapper check is just a best effort check for backward compatibility.

one_collect/src/helpers/dotnet/os/linux.rs
brianrob: It's technically possible, but right now we filter out non-executable mappings.  Are you thinking that you want to keep permissions to be a minimum (just read)?

brianrob: @beaubelgrave, do you have any concerns with just removing the protection check all-up for these call sites?  If needed, we can benchmark this, but I'm not too concerned.  The filename is already preloaded (not lazily) and many mappings will be file-backed and will start with a '/'.  I suspect that the filename checks are likely to bail pretty quickly if they aren't going to match.

beaubelgrave: > @beaubelgrave, do you have any concerns with just removing the protection check all-up for these call sites? If needed, we can benchmark this, but I'm not too concerned. The filename is already preloaded (not lazily) and many mappings will be file-backed and will start with a '/'. I suspect that the filename checks are likely to bail pretty quickly if they aren't going to match.

mmap's are not high volume events (except at startup). The CLR hook gets mmap events, even if they are not executable. So we can remove that check just for the DotNet part, so the impact I think should be manageable, especially if the mmap name is uncommon, IE: strcmp will bail very quickly within the string.

I'm not concerned with this until we have data indicating we need to do something different. I agree with @jkotas that if we keep the size the same in each release, we could use that if needed to get more perf.

PR #231, opened 2026-02-25
Perf: Drain session after disabling

beaubelgrave:
The universal session currently parses until told to stop, then disables the ring buffers. This can potentially skip data that was in the buffer(s) at the time it was told to stop.

Add a call to parse_all() after disabling the ring buffers to ensure all data within the buffers after disabling is parsed out. This ensures the output reflects all data up until the disable call.


brianrob (approved):
LGTM. Thanks.

PR #245, opened 2026-04-07
Fix OpenAt double-close and add perf map diagnostic logging

brianrob:
## Summary

Fixes a bug where `OpenAt` file descriptors were invalidated after clone due to a double-close, and adds diagnostic logging to help diagnose cases where JIT symbols are missing from perf maps.

## Bug Fix: OpenAt double-close on clone

`OpenAt` derived `Clone` but also implemented `Drop` to close its raw fd. When cloned, both copies shared the same fd, and dropping one would close it out from under the other, causing `EBADF` errors during perf-map file resolution. Fixed by replacing the raw fd with `Arc<OwnedFd>` so that clones share ownership and the fd is only closed when the last reference is dropped.

## Diagnostic Logging

Added warn/info/debug logging at critical points in the perf map and symbol resolution pipeline:

**Perf map enablement (dotnet helper):**
- `enable_perf_map`: warn on socket read/write errors and IPC result errors with pid/nspid context. Info on success.
- `disable_perf_map`: warn on socket write errors.
- `worker_thread_proc`: warn on PerfMapContext creation, environ read, and enable failures. Debug when skipping already-enabled processes.

**Symbol resolution (exporting):**
- `resolve_perf_map_symbols`: warn when perf-map file cannot be opened, with pid/ns_pid/error context.
- `add_matching_anon_symbols`: debug when anonymous mappings have no matching stack IPs.
- `add_matching_symbols`: info when 0 symbols matched a mapping.
- `add_matching_elf_symbols`: warn when root_fs unavailable. Debug for missing dev_node and empty symbol file results.

### DIFF
--- a/one_collect/src/helpers/dotnet/os/linux.rs
+++ b/one_collect/src/helpers/dotnet/os/linux.rs
@@ -145,17 +145,22 @@ impl PerfMapContext {
                     warn!("Failed to write to diagnostic socket: pid={}, nspid={}, error={}", self.pid, self.nspid, e);
                     anyhow::bail!("Failed to write to diagnostic socket: {}", e);
                 }
-                sock.read_exact(&mut result)?;
+                if let Err(e) = sock.read_exact(&mut result) {
+                    warn!("Failed to read diagnostic socket response: pid={}, nspid={}, error={}", self.pid, self.nspid, e);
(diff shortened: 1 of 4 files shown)


beaubelgrave (approved):
Nice! Thanks for tracking this done, looks good.

PR #248, opened 2026-04-09
ETW: Wire up Related/ActivityId to Exporting

### LINKED ISSUE #203: Support DotNet event Activity IDs
.NET events include [Activity IDs](https://learn.microsoft.com/en-us/dotnet/core/diagnostics/eventsource-activity-ids) which help correlate asynchronous work has it crosses different threads. For user_events the ActivityID and RelatedActivityID are written out in the [extension blob](https://github.com/dotnet/diagnostics/blob/main/documentation/design-docs/ipc-protocol.md#extension-blob-format) portion of the event binary data. For ETW ActivityID and RelatedActivityID are part of the events independent of .NET. The request is for OneCollect to read the IDs from the incoming events and include them in when it writes out event data to disk. In the NetTraceV6 format ActivityID and RelatedActivityID are stored in the [LabelListBlock](https://github.com/microsoft/perfview/blob/main/src/TraceEvent/EventPipe/NetTraceFormat.md#labellistblock) and an ID pointing to a LabelList row is stored in the event header.

Scenario showing Activity IDs working:
1. Create a default webapp: `dotnet new webapp`
2. Run the app: `dotnet run`
3. Trace some events with ActivityIds enabled: `dotnet-trace collect -n webapp3 --providers System.Threading.Tasks.TplEventSource:80:1,Microsoft.AspNetCore.Hosting:0:4`
4. Hit the webapp endpoint (its printed on the console when it started) to generate a few events
5. Open the trace in PerfView. In the Events view select Microsoft.AspNetCore.Hosting/Request/[Start|Stop] events and observe that ActivityID=<something> is present in the "Rest" column:

Image

PerfView does some manipulation so the value of the ID you see there isn't identical to the GUID that was present in the underlying file. To see that underlying GUID you can right-click on an event and select 'Dump Event' which will print some XML like this in the log:

```
<Event MSec= "21205.5629" PID="60612" PName="webapp3.exe"" TID="39968" ActivityID="00000014000000000000000005759d59" EventName="Request/Stop"
  TimeStamp="10/06/25 16:26:34.903562" ID="4" Version="0" Keywords="0x0000F00000000000" TimeStampQPC="22,818,628,405,356" QPCTime="0.100us"
  Level="Informational" ProviderName="Microsoft.AspNetCore.Hosting" ProviderGuid="9ded64a4-414c-5251-dcf7-1e4e20c15e70" ClassicProvider="False" ProcessorNumber="0"
  ActivityID="/#60612/1/4/" RawActivityID="00000014-0000-0000-0000-000005759d59"
  Opcode="2" Task="Default" Channel="0" PointerSize="8"
  CPU="0" EventIndex="9" TemplateType="DynamicTraceEventData">
  <PrettyPrint>
    <Event MSec= "21205.5629" PID="60612" PName="webapp3.exe"" TID="39968" ActivityID="00000014000000000000000005759d59" EventName="Request/Stop" ProviderName="Microsoft.AspNetCore.Hosting"/>
  </PrettyPrint>
  <Payload Length="0">
  </Payload>
</Event>
```

RawActivityID shows the value that was in the file.

This example didn't include RelatedActivityId because there was no nesting. To get an example that reliably nests, check out the source example in the demo here: [Activity IDs](https://learn.microsoft.com/en-us/dotnet/core/diagnostics/eventsource-activity-ids#tracking-work-using-an-activity-id). That demo also prints out the IDs to the console so you can compare that to what you see in OneCollect and in PerfView.

Repro:

Run either one of the demos as above, but instead of using dotnet-trace collect, use record-trace to capture the data. The goal is that the nettrace file produced by record trace includes the same Activity/Related Activity IDs that dotnet-trace would have produced.

beaubelgrave:
ETW supports ActivityId and RelatedActivityId for events on Windows. The exporting helper supports tracking both of these attributes by default. However, we have not wired up the ETW attributes to the exporting attributes.

Add necessary callbacks to read Related/ActivityId attributes from ETW Event structures when asked for by exporting.

### DIFF
--- a/one_collect/src/etw/mod.rs
+++ b/one_collect/src/etw/mod.rs
@@ -109,6 +109,19 @@ impl AncillaryData {
         }
     }
 
+    pub fn related_activity(&self) -> Option<Guid> {
+        if let Some(ext) = self.find_ext(
+            abi::EVENT_HEADER_EXT_TYPE_RELATED_ACTIVITYID) {
+            unsafe {
+                if (*ext).DataSize == 16 {
+                    return Some(*((*ext).DataPtr as *const Guid));
+                }
+            }
+        }
+
+        None
+    }
+
     pub fn op_code(&self) -> u8 {
         match self.event {
             Some(event) => {
--- a/one_collect/src/helpers/exporting/os/windows.rs
+++ b/one_collect/src/helpers/exporting/os/windows.rs
@@ -164,13 +164,16 @@ impl ExportSamplerOSHooks for ExportSampler {
     fn os_event_activity_id(
         &self,
         _data: &EventData) -> anyhow::Result<Option<[u8; 16]>> {
-        Ok(None)
+        Ok(Some(self.os.ancillary.borrow().activity().to_bytes()))
     }
 
     fn os_event_related_activity_id(
         &self,
         _data: &EventData) -> anyhow::Result<Option<[u8; 16]>> {
-        Ok(None)
+        Ok(match self.os.ancillary.borrow().related_activity() {
+            Some(id) => { Some(id.to_bytes()) },
+            None => { None },
+        })
     }
 
     fn os_event_callstack(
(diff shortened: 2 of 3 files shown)


brianrob (approved):
LGTM.  Thanks!

Issue #187, opened 2025-08-27
Support filter arguments


noahfalk:
In addition to provider id, keyword and level, ETW/EventPipe/EventSource all have a 4th kind of configuration data called arguments or filter arguments. For EventPipe this is a string that encodes arbitrary key-value pairs can be interpretted by a given provider. For example the LoggingEventSource uses arguments to encode a nested set of ILogger filters to determine which ILogger log messages should get emitted as EventSource events. https://learn.microsoft.com/en-us/aspnet/core/fundamentals/logging/?view=aspnetcore-9.0#event-source

For EventPipe in particular arguments are specified using the "arguments" field of the streaming_provider_config documented here: https://github.com/dotnet/diagnostics/blob/main/documentation/design-docs/ipc-protocol.md#collecttracing5

The minimal request is to support passing an arbitrary arguments string when enabling .NET events with a record-trace script. For example perhaps arguments are a new parameter to the event_from_dotnet() function:

let event = event_from_dotnet(
    "SomeProviderName",
    0x0, 4, "ArbitraryArgumentsStringGoesHere", 3, "EventName");

The more generalized request that could be part of this work item or a separate work item is to support enabling events on a provider by only specifying provider name, keywords, level, and arguments. The event names and IDs are never enumerated and can't necessarily be predicted in advance. We can decide on any appropriate script API but one example might be:

let scenario = new_dotnet_scenario();
scenario.enable_provider("SomeProviderName",
                         0x0, // keywords
                         4,   // level
                         "ArbitraryArgumentsStringGoesHere");
use_dotnet_scenario(scenario);

beaubelgrave:
Example script is:
set_dotnet_filter_args("System.Runtime", "EventCounterIntervalSec=1");

let flags = new_dotnet_provider_flags();
flags.with_callstacks();

record_dotnet_provider("System.Runtime", 0, 5, flags);

PR #250, opened 2026-04-10
ETW: Allow Provider Wide Events

beaubelgrave:
On DotNet for Linux, we allow recording provider level events using a keyword and level filter. On Windows we just had a TODO item there. We need to be able to record all events within a given provider that meet the keyword and level constraints.

Add new wildcard ID flag to Event, this flags if an Event should represent all events within a scope. For Windows the scope is an ETW provider. On Linux, there is no scope currently, and the flag is ignored.

Add OS abstraction to get the actual event ID. This allows us to get the event ID from Windows ETW sources. On Linux, this is only available on a subset of events, so we return None since there is already a pattern for this.

Use new wildcard ID flag and create a provider wide event on Windows that records all events coming in. All events will currently be treated as non-self describing events and will have empty format details. For well known providers with documented formats this will work well. For self describing events, such as TraceLogging events, these won't work currently.

Use new os_event_id() abstactions to mark the actual event ID within the record payloads instead of the wildcard event (which has an ID of 0).

Add callstack keyword filtering to ETW, utilize this when provider wide recording is enabled and a keyword comes in with not all the bits set (at that point it's not a filter).

This addresses most of #187, however, self describing events are still missing, so keeping it open.

### DIFF
--- a/one_collect/src/etw/abi.rs
+++ b/one_collect/src/etw/abi.rs
@@ -103,6 +103,7 @@ const PROCESS_TRACE_MODE_EVENT_RECORD: u32 = 268435456;
 const EVENT_FILTER_TYPE_PID: u32 = 0x80000004;
 const EVENT_FILTER_TYPE_EVENT_ID: u32 = 0x80000200;
 const EVENT_FILTER_TYPE_STACKWALK: u32 = 0x80001000;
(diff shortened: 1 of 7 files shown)


brianrob:
Apologies - I started looking at this and then promptly forgot.

brianrob (approved):
LGTM.  Thanks for implementing this.

PR #260, opened 2026-04-24
Fix Incomplete Call Stacks on Linux

### LINKED ISSUE #255: Incomplete stacks for .NET events on Linux
Stacks associated with .NET events (exceptions, allocs, etc.) are incomplete. For example, here's an exception stack (out of TraceEvent) for a process my team owns:

```
libcoreclr.so!DispatchManagedException(Object*, _CONTEXT*, _EXCEPTION_RECORD*)
libcoreclr.so!DispatchCallSimple(unsigned long*, unsigned int, unsigned long, unsigned int)
libcoreclr.so!CallDescrWorkerInternal
system.private.corelib!System.Runtime.EH::DispatchEx(valuetype System.Runtime.StackFrameIterator&, valuetype System.Runtime.EH/ExInfo&)
libcoreclr.so!SfiInit
libcoreclr.so!SfiInitWorker(StackFrameIterator*, _CONTEXT*, bool, bool*)
libcoreclr.so!ETW::ExceptionLog::ExceptionThrown(CrawlFrame*, int, int)
libcoreclr.so!EventPipeWriteEventExceptionThrown_V1
libcoreclr.so!ep_write_event(_EventPipeEvent*, unsigned char*, unsigned int, unsigned char const*, unsigned char const*)
libcoreclr.so!write_event_2(Thread*, _EventPipeEvent*, _EventPipeEventPayload*, unsigned char const*, unsigned char const*, Thread*, _EventPipeStackContents*)
libcoreclr.so!ep_session_write_event(_EventPipeSession*, Thread*, _EventPipeEvent*, _EventPipeEventPayload*, unsigned char const*, unsigned char const*, Thread*, _EventPipeStackContents*)
libc.so.6!writev
vmlinux!entry_SYSCALL_64_after_hwframe
vmlinux!do_syscall_64
vmlinux!x64_sys_call
vmlinux!__x64_sys_writev
vmlinux!do_writev
vmlinux!vfs_writev
vmlinux!do_iter_write
vmlinux!do_iter_readv_writev
vmlinux!user_events_write_iter
vmlinux!user_events_write_core.isra.0
```

I was able to get the full stack after hacking on record-trace locally:

```
libc.so.6!clone3
libc.so.6!start_thread
libcoreclr.so!CorUnix::CPalThread::ThreadEntry(void*)
libcoreclr.so!KickOffThread(void*)
libcoreclr.so!ManagedThreadBase::KickOff(void (*)(void*), ManagedThreadBase::KickOff)
libcoreclr.so!KickOffThread_Worker(void*)
libcoreclr.so!DispatchCallSimple(unsigned long*, unsigned int, unsigned long, unsigned int)
libcoreclr.so!CallDescrWorkerInternal
system.private.corelib!System.Threading.Thread::StartCallback()
...
<managed-frames>
...
system.private.corelib!System.Runtime.ExceptionServices.ExceptionDispatchInfo::Throw()
libcoreclr.so!IL_Throw(Object*)
libcoreclr.so!DispatchManagedException(Object*, _CONTEXT*, _EXCEPTION_RECORD*)
libcoreclr.so!DispatchCallSimple(unsigned long*, unsigned int, unsigned long, unsigned int)
libcoreclr.so!CallDescrWorkerInternal
system.private.corelib!System.Runtime.EH::DispatchEx(valuetype System.Runtime.StackFrameIterator&, valuetype System.Runtime.EH/ExInfo&)
libcoreclr.so!SfiInit
libcoreclr.so!SfiInitWorker(StackFrameIterator*, _CONTEXT*, bool, bool*)
libcoreclr.so!ETW::ExceptionLog::ExceptionThrown(CrawlFrame*, int, int)
libcoreclr.so!EventPipeWriteEventExceptionThrown_V1
libcoreclr.so!ep_write_event(_EventPipeEvent*, unsigned char*, unsigned int, unsigned char const*, unsigned char const*)
libcoreclr.so!write_event_2(Thread*, _EventPipeEvent*, _EventPipeEventPayload*, unsigned char const*, unsigned char const*, Thread*, _EventPipeStackContents*)
libcoreclr.so!ep_session_write_event(_EventPipeSession*, Thread*, _EventPipeEvent*, _EventPipeEventPayload*, unsigned char const*, unsigned char const*, Thread*, _EventPipeStackContents*)
libc.so.6!writev
vmlinux!entry_SYSCALL_64_after_hwframe
vmlinux!do_syscall_64
vmlinux!x64_sys_call
vmlinux!__x64_sys_writev
vmlinux!do_writev
vmlinux!vfs_writev
vmlinux!do_iter_write
vmlinux!do_iter_readv_writev
vmlinux!user_events_write_iter
vmlinux!user_events_write_core.isra.0
```

I'm sure my patch (attached) isn't right, but it proves the full stack is there. I've only looked at stacks from exception and memory alloc events. I imagine other .NET event types have the same problem.

I don't think you need anything special to repro. Just a .NET app that throws and allocates. I collected traces with

```
record-trace --pid <pid> --on-cpu --script-file <file>
```

and the script

```
let scenario = new_dotnet_scenario();
scenario.with_records(); // records actual data
scenario.with_callstacks(); // records callstacks
scenario.with_gc_allocs(); // enables GC alloc events
scenario.with_exceptions(); // enables Exception events
use_dotnet_scenario(scenario);
```

[0001-.NET-event-stack-hacks.patch](https://github.com/user-attachments/files/26788381/0001-.NET-event-stack-hacks.patch)

brianrob:
## Problem
On Linux, callstacks captured for some processes were truncated. Symptoms ranged from stacks ending after one or two frames to deep cross-ABI unwinds terminating prematurely. The issue was particularly visible on .NET workloads (libcoreclr.so happens to exercise all of the affected code paths at once), but the underlying bugs are general and can affect any binary.

## Root causes
Three independent bugs in the unwind/lookup pipeline conspired:

1. **Wrong VA used for `.eh_frame` / `.eh_frame_hdr` lookups.** ELF binaries can have a non-zero `(p_vaddr - p_offset)` delta on their executable `PT_LOAD` segment. one-collect was treating the in-mapping offset as a VA directly, so DWARF FDE binary searches missed the correct entry for any binary with such a delta. libcoreclr.so is one prominent example (delta `~0x1000`), but the fix applies to every ELF.
2. **`.eh_frame_hdr` binary search missed an upper-bound check.** Even when the VA was right, the lookup could return an FDE whose PC range did not actually cover the IP, causing the unwinder to apply the wrong rules and corrupt the next frame.
3. **x64 prolog scanner gave up too early on cross-ABI transitions.** When unwinding across stub trampolines that have no DWARF coverage, the scanner needed slightly more recovery latitude to keep walking.

A secondary issue: the default DWARF user-stack capture (8 KiB) was too small for the deepest stacks we observed, so even correct unwinds were truncated by missing stack bytes.

## Changes
Five commits, each independently reviewable:

| Commit | Summary |
|---|---|
| `1baf009` | ruwind: enforce FDE PC upper bound in `.eh_frame_hdr` lookup |
| `a71010b` | ruwind: improve x64 prolog/scan recovery for cross-ABI unwinds |
| `72c851c` | callstack/linux: bump default DWARF user-stack capture from 8 KiB to 32 KiB |
| `deeae72` | Translate IPs to ELF VAs using executable PT_LOAD delta (issue #255) |
| `f2deff4` | ruwind: add x86_64 unwinder unit tests |

## Validation
- `cargo test`: ruwind 11/11, one_collect 76/76 — all passing.
- New x86_64 unwinder unit tests covering the FDE upper-bound and prolog-scan cases.
- End-to-end captures against a .NET 10 repro app exercising allocation, GC, contention, and exception paths:
  - **on-cpu**: 20,975 samples, depth 1–63, full chain through JIT, runtime, managed code, host, libc, and `[stack]`.
  - **Exception/Start** (80,228 events, depth 45), **GC/AllocationTick** (depth 49), **Contention/Start** (depth 33) — all walk cleanly from the kernel `writev` syscall through the runtime's EventPipe, into managed methods, and back out to `__libc_start_main`.
- Native (non-.NET) workloads (`bash`, `seq`, `find`/`grep`) unwind to depths 21–26 with full kernel + libc + user resolution — no regression.
- Local symbol resolution unaffected: runtime exports continue to resolve correctly during capture.

Fixes #255.

### DIFF
--- a/one_collect/src/helpers/callstack/os/linux.rs
+++ b/one_collect/src/helpers/callstack/os/linux.rs
@@ -3,7 +3,6 @@
 
 use std::fs::File;
 use std::path::PathBuf;
-use std::os::fd::{FromRawFd, IntoRawFd, RawFd};
 use std::ops::DerefMut;
 use std::collections::HashMap;
 use std::collections::hash_map::Entry::{self, Vacant};
@@ -13,10 +12,11 @@ use super::*;
 use crate::PathBufInteger;
 use crate::perf_event::*;
 use crate::event::DataFieldRef;
+use crate::openat::DupFd;
 use crate::Writable;
 
-use libc::*;
 use ruwind::*;
+use libc::PROT_EXEC;
 
 pub struct CallstackReader {
     state: Writable<MachineState>,
@@ -31,20 +31,55 @@ impl Clone for CallstackReader {
 }
 
 struct ModuleLookup {
-    fds: HashMap<ModuleKey, RawFd>,
+    fds: HashMap<ModuleKey, DupFd>,
+    va_offsets: HashMap<ModuleKey, u64>,
 }
 
 impl ModuleLookup {
     fn new() -> Self {
         Self {
             fds: HashMap::new(),
+            va_offsets: HashMap::new(),
         }
     }
 
     fn entry(
         &mut self,
-        key: ModuleKey) -> Entry<'_, ModuleKey, RawFd> {
+        key: ModuleKey) -> Entry<'_, ModuleKey, DupFd> {
         self.fds.entry(key)
+    }
+
+    fn va_offset(
+        &mut self,
+        key: ModuleKey) -> u64 {
+        if let Some(value) = self.va_offsets.get(&key) {
+            return *value;
+        }
+
+        let mut value = 0u64;
+        /*
+         * Read the load header from the cached FD rather than re-opening
+         * by path so we cannot race with a file that has been replaced
+         * out from under us between the original open and now.
+         */
+        if let Some(fd) = self.fds.get(&key) {
+            match fd.open() {
+                Some(file) => {
+                    let mut reader = std::io::BufReader::new(file);
+                    if let Ok(load_header) = ruwind::elf::get_load_header(&mut reader) {
+                        value = load_header
+                            .p_vaddr()
+                            .saturating_sub(load_header.p_offset());
+                    }
+                },
+                None => {
+                    warn!("Failed to dup module FD for va_offset lookup: dev={}, ino={}", key.dev(), key.ino());
+                },
+            }
+        }
+
+        self.va_offsets.insert(key, value);
+        value
     }
 }
 
@@ -52,14 +87,8 @@ impl ModuleAccessor for ModuleLookup {
     fn open(
         &self,
         key: &ModuleKey) -> Option<File> {
-        match self.fds.get(&key) {
-            Some(fd) => {
-                /* Clone it and return for caller */
-                unsafe {
-                    let cloned_fd = dup(*fd);
-                    Some(File::from_raw_fd(cloned_fd))
-                }
-            },
+        match self.fds.get(key) {
+            Some(fd) => { fd.open() },
             None => { None },
         }
     }
@@ -144,21 +173,27 @@ impl MachineState {
            filename.starts_with("/memfd:") ||
            filename.starts_with("//anon");
 
-        if !mem_backed {
-            /* File backed */
-            let key = ModuleKey::new(dev, ino);
+        let key = if !mem_backed {
+            Some(ModuleKey::new(dev, ino))
+        } else {
+            None
+        };
+
+        if let Some(key) = key {
+            /* File backed - build the /proc/<pid>/root/<filename> path
+             * once and share it between the FD cache and the va_offset
+             * lookup so we don't construct it twice. */
+            self.path.clear();
+            self.path.push("/proc");
+            self.path.push_u32(pid);
+            self.path.push("root");
+            self.path.push(filename);
 
             if let Vacant(entry) = self.modules.entry(key) {
                 /* Try to open and keep a single FD for that file */
-                self.path.clear();
-                self.path.push("/proc");
-                self.path.push_u32(pid);
-                self.path.push("root");
-                self.path.push(filename);
-
                 /* Only insert if we can actually open it */
                 if let Ok(file) = std::fs::File::open(&self.path) {
-                    entry.insert(file.into_raw_fd());
+                    entry.insert(DupFd::new(file));
                 } else {
                     warn!("Failed to open module file: pid={}, filename={}", pid, filename);
                 }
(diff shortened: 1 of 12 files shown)


one_collect/src/helpers/callstack/os/linux.rs
```diff
@@ -386,7 +386,7 @@ impl CallstackHelper {
             unwinder: None,
             external_lookup: false,
             ip_only: false,
-            stack_size: 4096,
+            stack_size: 32768,
```
beaubelgrave: That's a spicy meatball... Are we sure we want 32KB for all callers of one-collect by default? I can see some tools having a higher default, but that's quite high. In other tools I've worked on in the past we have 8KB as the default size.

one_collect/src/helpers/callstack/os/linux.rs
```diff
@@ -46,6 +48,38 @@ impl ModuleLookup {
...
+                let cloned_fd = dup(*fd);
+                if cloned_fd >= 0 {
+                    let file = File::from_raw_fd(cloned_fd);
+                    let mut reader = std::io::BufReader::new(file);
+                    if let Ok(load_header) = ruwind::elf::get_load_header(&mut reader) {
+                        value = load_header
+                            .p_vaddr()
+                            .saturating_sub(load_header.p_offset());
+                    }
+                }
```
beaubelgrave: Probably should trace here that the dup failed?

beaubelgrave: Can we just use the DupFd we have as well?

beaubelgrave (approved):
I approve these changes, this is a TON of work, nicely done!

I do want us to have a conversation about the 32KB being default, I'm fine with it right now, but I suspect that is likely too high to be default out of the box for all solutions, but I don't know what the IDEAL value is either.

Other comments are things to go do afterwards and nice to haves.

one_collect/src/helpers/callstack/os/linux.rs
brianrob: That's a fair point - I just pushed a change that only switches to 32kB if we're capturing dotnet events.  How does that sound?

one_collect/src/helpers/callstack/os/linux.rs
brianrob: Good call.  Updated.

brianrob:
> I approve these changes, this is a TON of work, nicely done!
> 
> I do want us to have a conversation about the 32KB being default, I'm fine with it right now, but I suspect that is likely too high to be default out of the box for all solutions, but I don't know what the IDEAL value is either.
> 
> Other comments are things to go do afterwards and nice to haves.

Thanks much! I figured I'd see if I can make the changes here since they didn't seem too bad.  PTAL again. Thanks.

beaubelgrave (approved):
Looks good, thanks! At some point we may have to make the dotnet helper configurable for stack size (or do some scaler). But I don't think that time is now, maybe when we visit SFRAME stuff.

brianrob:
> Looks good, thanks! At some point we may have to make the dotnet helper configurable for stack size (or do some scaler). But I don't think that time is now, maybe when we visit SFRAME stuff.

Thanks.  Yeah, sounds great.

PR #267, opened 2026-05-11
Make event record publicly accessible

swashtek:
Make event record publicly accessible for TDH decoding

### DIFF
--- a/one_collect/src/etw/mod.rs
+++ b/one_collect/src/etw/mod.rs
@@ -82,6 +82,13 @@ impl AncillaryData {
         }
     }
 
+    /// Exposes the raw EVENT_RECORD pointer for TDH decoding.
+    /// Returns None if no event is currently being processed.
+    pub fn raw_event_record(&self) -> Option<*const EVENT_RECORD> {
+        self.event
+    }
+
+
     pub fn time(&self) -> u64 {
         match self.event {
             Some(event) => {
--- a/one_collect/src/etw/mod.rs
+++ b/one_collect/src/etw/mod.rs
@@ -19,10 +19,11 @@ mod events;
 use abi::{
     TraceSession,
     TraceEnable,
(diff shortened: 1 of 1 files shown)


brianrob:
Thanks @swashtek for contributing this.  Can you share a bit more about what you're looking to do here?  My first thought is that this new callback will get called for every event regardless of the user's configuration - there is filtering that happens after this.

cc: @beaubelgrave

one_collect/src/etw/mod.rs
```diff
@@ -82,6 +83,13 @@ impl AncillaryData {
         }
     }
 
+    /// Exposes the raw EVENT_RECORD pointer for TDH decoding.
+    /// Returns None if no event is currently being processed.
+    pub fn raw_event_record(&self) -> Option<*const EVENT_RECORD> {
```
beaubelgrave, line 88: You can utilize the provider wide callback mechanism that already exists for this, right?

beaubelgrave: There are per-OS event extensions to get the underlying data. We should use that for the EVENT_RECORD access along with the provider wide callback mechanisms. This aligns well with our OS abstraction philosophy at the Event level.

beaubelgrave: See https://github.com/microsoft/one-collect/blob/main/one_collect/src/event/os/windows.rs for where to add EVENT_RECORD access from a normal Event struct for Windows.

utpilla:
We may not require marking `EVENT_RECORD` public if we go forward with the approach mentioned in #270.

one_collect/src/etw/mod.rs
swashtek: Thank you for the feedback. I will abandon this change and follow the approach mentioned in #270.

Issue #270, opened 2026-05-13
TDH Decoding Support for Tracelogging and TraceloggingDynamic ETW events


utpilla:
## Summary

This proposal is a high level design for adding support for decoding arbitrary **TraceLogging** and **TraceLoggingDynamic** ETW events using the Windows [Trace Data Helper (TDH)](https://learn.microsoft.com/en-us/windows/win32/etw/retrieving-event-data-using-tdh) APIs. Today, the ETW layer in `one_collect` can only consume events whose field layouts are hardcoded at compile time. This proposal extends the system to handle self-describing events where the schema is discovered at runtime, while reusing the existing `EventFormat` / `EventData` / `Event::process()` pipeline.

Manifest-based event decoding is explicitly deferred but the design accommodates it as a future extension.

## Motivation

The current ETW implementation supports a fixed set of well-known kernel events (process, image load, CPU sample, context switch, DPC, page fault, callstack) and custom provider events where the caller supplies the field layout upfront. Each event's schema is manually declared in [`etw/events.rs`](https://github.com/microsoft/one-collect/blob/main/one_collect/src/etw/events.rs) with hardcoded offsets:

```rust
// Current approach - manual offset arithmetic per field
format.add_field(EventField::new(
    "ProcessId".into(), "u32".into(),
    LocationType::Static, offset, len));
offset += len;
```

This does not work for **TraceLogging** or **TraceLoggingDynamic** providers, where the event schema is embedded in the event itself (via `EVENT_HEADER_EXT_TYPE_EVENT_SCHEMA_TL`) and varies per event type. To consume these events, we need to discover the schema at runtime.

The TDH APIs ([`TdhGetEventInformation`](https://docs.rs/windows-sys/0.61.2/windows_sys/Win32/System/Diagnostics/Etw/fn.TdhGetEventInformation.html)) decode any `EVENT_RECORD` into a `TRACE_EVENT_INFO` structure describing the event's properties - their names, types, sizes, and relationships. This is the standard Windows mechanism for decoding self-describing ETW events.

TraceLogging is widely used across Windows itself and across applications that target Windows. Decode support lets `one_collect` consume any of these through one code path - whether the consumer wants every field of every event (e.g. the [opentelemetry-etw-logs](https://github.com/open-telemetry/opentelemetry-rust-contrib/tree/main/opentelemetry-etw-logs) consumption side) or just a few specific fields from a specific event.

### In scope
- Decoding TraceLogging and TraceLoggingDynamic events
[...]

PR #273, opened 2026-05-19
Use types from windows-sys

utpilla:
### Changes

Replaces four hand-rolled ETW types that surface (or will surface) on the crate's public API with re-exports from `windows-sys`:

- `EVENT_RECORD`
- `EVENT_HEADER`
- `EVENT_HEADER_EXTENDED_DATA_ITEM`
- `EVENT_DESCRIPTOR`

### Why

Precursor to the TDH decoder work tracked in #270. That decoder will take a `&EVENT_RECORD` from `AncillaryData::record()` and hand it to a TDH-based decode call. Sourcing the public types from `windows-sys` gives us a single type identity end to end and shrinks the `one_collect` crate's public API surface.

This was the open question I raised on #270, and this PR is following [Option B mentioned in the issue](https://github.com/microsoft/one-collect/issues/270#issuecomment-4483260376).

### Scope

Deliberately constrained to only the four public types. The rest of `etw/abi.rs` stays hand-rolled:

- All `extern "system"` blocks (`StartTraceW`, `ControlTraceW`, `EnableTraceEx2`, `OpenTraceW`, `ProcessTrace`, `CloseTrace`, `TraceQueryInformation`, `TraceSetInformation`, plus the privilege helpers and `GetActiveProcessorCount`).
- All internal-only structs (`EVENT_TRACE_PROPERTIES`, `WNODE_HEADER`, `ENABLE_TRACE_PARAMETERS`, `EVENT_FILTER_DESCRIPTOR`, `EVENT_TRACE_LOGFILE`, `TRACE_LOGFILE_HEADER`, `EVENT_TRACE`, `EVENT_TRACE_HEADER`, `TRACE_PROFILE_INTERVAL`, `TOKEN_PRIVILEGES`, `EVENT_FILTER_LEVEL_KW`, `CLASSIC_EVENT_ID`).
- All `TRACE_LEVEL_*` / `EVENT_CONTROL_CODE_*` / `EVENT_HEADER_EXT_TYPE_*` / `EVENT_ENABLE_PROPERTY_*` constants.
- `TraceSession`, `TraceEnable`, `TraceFilter`, `flush_trace`, `wide_string`.

A full migration of the FFI is a fair follow-up but out of scope here to keep this PR focused on the bare-minimum change the TDH work needs.

### Safety

The trait impl contains three `unsafe` operations, each backed by a `const _: () = assert!(...)` guard above the impl block so future `windows-sys` layout drift fails the build instead of producing wrong bytes silently:

- `provider_guid` / `activity_guid` transmute `windows_sys::core::GUID` → our `Guid`. Guarded by `size_of` + `align_of` equality asserts.
- `processor_index` reads `u16` out of a union arm. Guarded by an asserted `size_of == 2` and an `align_of >= align_of::<u16>()` on `ETW_BUFFER_CONTEXT_0`.

### Follow-ups (not in this PR)

- The TDH decoder itself (#270).
- Optional: migrate the rest of `etw/abi.rs`'s FFI to `windows-sys` in a focused later PR if we ever want to retire the hand-rolled extern blocks.

### DIFF
--- a/one_collect/Cargo.lock
+++ b/one_collect/Cargo.lock
@@ -491,6 +491,7 @@ dependencies = [
  "sha1",
  "tracing",
  "twox-hash",
+ "windows-sys",
  "winreg",
 ]
 
--- a/THIRD-PARTY-NOTICES.md
+++ b/THIRD-PARTY-NOTICES.md
@@ -757,3 +757,35 @@ CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
 OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR
 IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
 DEALINGS IN THE SOFTWARE.
+
+
+windows-sys
+===========
+
+MIT License
+
+Copyright (c) Microsoft Corporation.
+
+Permission is hereby granted, free of charge, to any
+person obtaining a copy of this software and associated
+documentation files (the "Software"), to deal in the
+Software without restriction, including without
+limitation the rights to use, copy, modify, merge,
+publish, distribute, sublicense, and/or sell copies of
+the Software, and to permit persons to whom the Software
+is furnished to do so, subject to the following
+conditions:
+
+The above copyright notice and this permission notice
+shall be included in all copies or substantial portions
+of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF
+ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED
+TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A
+PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT
+SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY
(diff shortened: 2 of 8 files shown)


one_collect/Cargo.lock
```diff
@@ -491,6 +491,7 @@ dependencies = [
  "sha1",
  "tracing",
  "twox-hash",
+ "windows-sys",
```
beaubelgrave, line 494: @brianrob, do we need to list this in third parties md? Not sure if this is technically 3rd party or not.

beaubelgrave (approved):
Approved, looks good, thanks for doing this! We may require a follow up item on the third party tracking of windows-sys if it's technically 3rd party.

beaubelgrave:
Windows-sys uses MIT license, we should go ahead and add it to our THIRD-PARTY-NOTICES.md before merging. If you update this PR with that fix, I'll go ahead and approve again and merge. @utpilla.

utpilla:
> Windows-sys uses MIT license, we should go ahead and add it to our THIRD-PARTY-NOTICES.md before merging. If you update this PR with that fix, I'll go ahead and approve again and merge. @utpilla.

I have updated `THIRD_PARTY_NOTICES.md`.

Reference: https://github.com/microsoft/windows-rs/blob/master/license-mit

beaubelgrave (approved):
Thanks!

Issue #254, opened 2026-04-15
[Linux] Support wakeup watermark for perf ring buffers


lalitb:
On Linux, one-collect currently polls the perf ring buffer on a fixed sleep interval when no data is available. The perf subsystem supports event-driven wakeup via a configurable byte watermark - the kernel signals the perf fd only when enough data has accumulated, eliminating unnecessary polling.

  Supporting this would allow callers to tune the latency/throughput tradeoff:

  - **Low watermark** -> wake on every event, minimum latency
  - **High watermark** -> amortize wakeup cost, better throughput at scale

beaubelgrave:
Lets do this as an option during the builder pattern(s) we have.

lalitb:
Sounds good. I’ll wire it up as an opt-in builder option, leaving existing polling behavior unchanged by default.

utpilla:
There are two levels of polling to address here.

**Level 1: Kernel level**

Wire up `wakeup_watermark` so the kernel signals the perf fd when data arrives. This is a prerequisite for level 2.

**Level 2: Consumer-level readiness**

Even with watermarks wired up, consumers have no way to receive the kernel's readiness signal. `one_collect`'s API is pull-based. You call `parse_for_duration`, it reads whatever is in the ring and returns. There's no "notify me when data is available" mechanism.

For consumers that run on async runtimes (e.g., a single-threaded tokio `current_thread` runtime) neither blocking nor polling is viable. They need a fd they can register with their own event loop.

Proposed API: expose per-CPU perf fds so consumers can decide how to compose them, single-core consumers use one fd directly, multi-core consumers create their own epoll:

Option A lookup by CPU:

```rust
impl PerfSession {
    pub fn perf_fd_for_cpu(&self, cpu: u32) -> Option<BorrowedFd<'_>>;
}
```
Option B iterate all:

```rust
impl PerfSession {
    pub fn perf_fds(&self) -> impl Iterator<Item = (u32, BorrowedFd<'_>)>; // could return a slice as well?
}
```
Option C both, for convenience.

In all cases, `BorrowedFd<'_>` ties the fd lifetime to the session, preventing use-after-close. **`one_collect` retains ownership and stays runtime-agnostic** (no dependency on any particular async runtime implementation). Consumers build their own async wrappers (e.g., [`AsyncFd`](https://docs.rs/tokio/latest/tokio/io/unix/struct.AsyncFd.html) on top.

[...]

PR #276, opened 2026-05-20
Support wakeup watermark for perf ring buffers

utpilla:
Closes part of #254 (first of two PRs; the follow-up will expose the per-CPU perf fds so a consumer can actually wait on them via `epoll(2)` / `tokio::io::unix::AsyncFd`).

## Changes

Adds an opt-in `RingBufSessionBuilder::with_wakeup_watermark(bytes: u32)` that configures the kernel to defer marking each per-CPU perf fd readable until at least `bytes` of data have accumulated in its ring buffer.

**Motivation**

The kernel default is to mark the perf fd readable on every event. For a future epoll / `AsyncFd`-driven consumer this means one wakeup per record, which is expensive under load. The watermark lets the caller amortize wakeup cost against latency.

- New helper `ring_data_bytes(page_count)` in `perf_event/rb/mod.rs` computes the per-CPU ring data area in bytes (`next_power_of_two(page_count) * PAGE_SIZE`). This is the actual bound the watermark must stay under; making it a named function gives us a single source of truth and lets the unit tests reference exactly the same value the builder uses.
- New `RingBufBuilder<T>::set_wakeup_watermark(bytes)` (crate-visible) sets `FLAG_WATERMARK` and `wakeup_events_watermark` on the underlying `perf_event_attr`.
- New public `RingBufSessionBuilder::with_wakeup_watermark(bytes)` stores the requested value on the builder. It is threaded through every existing rebuild block alongside `pages` / `target_pids` / `target_cpus`.
- `RingBufSessionBuilder::build()` applies the watermark to the kernel/leader `perf_event_attr` only. The leader fd is the one the kernel wakes when data lands in its per-CPU ring; all other event types (tracepoint, profiling, context-switch, page-fault, BPF) redirect their output into the same ring via `PERF_EVENT_IOC_SET_OUTPUT`, so the leader's watermark covers them.
- `build()` validates that `bytes < ring_data_bytes(self.pages)` and returns `Err(io_error(...))` otherwise. This is necessary because the kernel does not validate this at `perf_event_open()` time and silently never wakes the fd if the watermark cannot be reached: the wakeup check is `head - rb->wakeup > rb->watermark`, and `head - rb->wakeup` is bounded above by the data area size, so a watermark `>=` that size can never trigger a wakeup and a consumer waiting on the fd would stall while the kernel drops records as `PERF_RECORD_LOST`.
- `CpuRingBuf::create_reader` is refactored to use the new helper (`ring_data_bytes(page_count) + page_size` for the mmap length, the `+ page_size` being the perf metadata page). The previous expression `page_count.next_power_of_two() + 1` (pages, with the metadata page folded into the count) computes the same total length; the refactor just expresses it in terms of the shared helper.

## Tests

Three unit tests on the boundary, all using `ring_data_bytes(pages)` directly so they are correct on any page size:

- `wakeup_watermark_rejected_when_equal_to_ring_size` - `bytes == ring_data_bytes` must error.
- `wakeup_watermark_rejected_when_above_ring_size` - `bytes == ring_data_bytes + 1` must error.
- `wakeup_watermark_accepted_when_below_ring_size` - `bytes == ring_data_bytes - 1` must pass builder validation (the subsequent `perf_event_open()` may still fail in test environments without the relevant privileges; the test asserts only that the failure, if any, is not the watermark validation error).

### DIFF
--- a/one_collect/src/perf_event/rb/mod.rs
+++ b/one_collect/src/perf_event/rb/mod.rs
@@ -111,6 +111,21 @@ pub fn cpu_count() -> u32 {
     }
 }
 
+/// Size in bytes of the per-CPU ring's data area for a user-supplied page
+/// count.
+///
+/// The kernel requires the data area to be a power-of-two number of
+/// pages, so the user's `page_count` is rounded up via
+/// `next_power_of_two()`. The actual `mmap` is one additional metadata
+/// page on top of this (`ring_data_bytes(n) + PAGE_SIZE`).
+///
+/// This is the bound the `wakeup_events_watermark` must be strictly
+/// smaller than for the kernel to ever wake the perf fd.
+pub(super) fn ring_data_bytes(page_count: usize) -> usize {
+    let page_size = unsafe { sysconf(_SC_PAGE_SIZE) as usize };
+    page_count.next_power_of_two() * page_size
+}
+
 pub fn perf_timestamp(
     attr: &perf_event_attr) -> u64 {
     unsafe {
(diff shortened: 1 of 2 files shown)


beaubelgrave (approved):
Looks great!

PR #278, opened 2026-05-27
Add TDH-based dynamic event decoder for TraceLogging events

swashtek:
### Description

Adds `TdhDecoder` — a runtime schema decoder for TraceLogging and TraceLoggingDynamic ETW events using the Windows TDH (Trace Data Helper) APIs. This enables `one_collect` to consume any TraceLogging provider without predefining event schemas at compile time.

#### What's new

- __`one_collect/src/etw/tdh.rs`__ — New module implementing:

  - `TdhDecoder` struct with schema caching (keyed by raw TL bytes, split by 32/64-bit pointer width)
  - Per-event offset computation that correctly handles variable-length fields (strings, binary) interleaved with fixed-size fields
  - `event_name()` accessor for TraceLogging event name identity
  - 17 unit tests covering scanning, offset computation, UTF-16 decoding, and type mapping

- __`one_collect/src/etw/mod.rs`__ — Updated to:

  - Remove `mod schema` (replaced by internal `SchemaLayoutCache` in `tdh.rs`)
  - Re-export `EVENT_RECORD` for downstream consumers of `TdhDecoder`
  - Expose `AncillaryData::raw_event_record()` for TDH decoding access

- __`one_collect/Cargo.toml`__ — Added `twox-hash` dependency for XxHash64 (matching existing ETW module usage)

#### Design decisions

1. __Two-tier caching__: Schema metadata (field names, types, `PropertyKind`) is cached permanently. For all-fixed-size schemas, a pre-built `EventFormat` is also cached (zero per-event allocation). For schemas with variable-length fields, concrete offsets are computed per-event by walking the actual payload.

2. __Aligned TDH buffer__: `AlignedTeiBuf` uses `Vec<u64>` instead of `Vec<u8>` to guarantee 8-byte alignment for `TRACE_EVENT_INFO` and `EVENT_PROPERTY_INFO` structs, eliminating undefined behavior from misaligned pointer casts.

3. __`PROPERTY_PARAM_LENGTH` guard__: When `PROPERTY_HAS_LENGTH` is combined with `PROPERTY_PARAM_LENGTH` (0x10), the length field is a property *index* (not a literal byte count). The decoder correctly skips the fixed-size optimization in this case.

4. __Zero-allocation type names__: `CachedProperty::type_name` uses `&'static str` instead of `String`, eliminating heap allocations for the fixed set of type name strings.

#### Not yet supported (future work)

- Manifest-based event decoding (returns `TdhDecodeError::NotFound`)
- Array-typed properties (placeholder `unsupported_array` field)
- Map/enum value resolution
- `tracing` instrumentation for cache metrics

### DIFF
--- a/one_collect/src/etw/mod.rs
+++ b/one_collect/src/etw/mod.rs
@@ -17,11 +17,8 @@ mod abi;
 mod events;
 pub mod tdh;
 
-/// Re-export `EVENT_RECORD` so consumers of `TdhDecoder` can name the
-/// type without reaching into `abi` (which is `pub(crate)`).
-pub use abi::EVENT_RECORD;
-
 use abi::{
+    EVENT_RECORD,
     TraceSession,
     TraceEnable,
     EVENT_HEADER_EXTENDED_DATA_ITEM,
@@ -87,12 +84,12 @@ impl AncillaryData {
         }
     }
 
-    /// Exposes the raw EVENT_RECORD pointer for TDH decoding.
-    /// Returns None if no event is currently being processed.
-    pub fn raw_event_record(&self) -> Option<*const EVENT_RECORD> {
-        self.event
+    /// Returns a borrow of the current `EVENT_RECORD`, or `None` if no
+    /// event is currently being processed.  The borrow is valid for the
+    /// duration of the ETW callback (same as `&self`).
+    pub fn record(&self) -> Option<&EVENT_RECORD> {
+        self.event.map(|p| unsafe { &*p })
     }
-
 
     pub fn time(&self) -> u64 {
         match self.event {
@@ -217,21 +214,10 @@ impl AncillaryData {
         ext_type: u32) -> Option<*const EVENT_HEADER_EXTENDED_DATA_ITEM> {
         match self.event {
             Some(event) => {
-                unsafe {
-                    let ext = (*event).ExtendedData;
-
-                    for i in 0..(*event).ExtendedDataCount as usize {
-                        let item = ext.add(i);
-
-                        if (*item).ExtType == ext_type as u16 {
-                            return Some(item);
-                        }
-                    }
-
-                    None
-                }
+                // Delegate to the centralized EventRecordExt method.
+                unsafe { (*event).find_extended_data(ext_type as u16) }
             },
-            None => { None},
+            None => None,
         }
     }
 }
--- a/one_collect/src/etw/mod.rs
+++ b/one_collect/src/etw/mod.rs
@@ -82,6 +82,13 @@ impl AncillaryData {
         }
     }
 
+    /// Exposes the raw EVENT_RECORD pointer for TDH decoding.
+    /// Returns None if no event is currently being processed.
+    pub fn raw_event_record(&self) -> Option<*const EVENT_RECORD> {
+        self.event
+    }
+
+
     pub fn time(&self) -> u64 {
         match self.event {
             Some(event) => {
--- a/one_collect/src/etw/mod.rs
+++ b/one_collect/src/etw/mod.rs
@@ -15,11 +15,15 @@ use crate::Guid;
 #[allow(dead_code)]
 mod abi;
 mod events;
+pub mod tdh;
+
+/// Re-export `EVENT_RECORD` so consumers of `TdhDecoder` can name the
+/// type without reaching into `abi` (which is `pub(crate)`).
+pub use abi::EVENT_RECORD;
 
 use abi::{
     TraceSession,
     TraceEnable,
-    EVENT_RECORD,
     EVENT_HEADER_EXTENDED_DATA_ITEM,
     CLASSIC_EVENT_ID,
     EventRecordExt,
--- a/one_collect/src/etw/tdh.rs
+++ b/one_collect/src/etw/tdh.rs
@@ -136,6 +136,8 @@ struct CachedSchema {
     /// fixed-size fields, and the framework's skip chain handles
     /// variable-length fields lazily via `size = 0`.
     format: EventFormat,
+    /// Monotonic ID assigned at insertion time.
+    schema_id: SchemaId,
 }
 
 /// Hash builder using XxHash64, matching the rest of the ETW module.
@@ -170,10 +172,9 @@ impl SchemaCache {
 
 /// Result of a successful [`TdhDecoder::decode`] call.
 ///
-/// Wraps the decoded [`EventData`] together with a flag indicating
-/// whether this was the first time the schema was seen (cache miss).
-/// Callers can use `is_new_schema` to register the [`EventFormat`]
-/// with an exporter exactly once.
+/// Contains the decoded [`EventData`], the TraceLogging `event_name`,
+/// and a monotonic [`SchemaId`] that exporters can use as a cheap
+/// lookup key for format registration.
 #[non_exhaustive]
 pub struct TdhDecodedEvent<'a> {
     /// The decoded event data.
@@ -184,11 +185,28 @@ pub struct TdhDecodedEvent<'a> {
     /// to the event ID, which is often 0).  Consumers can use this for
     /// OTEL log record naming without a second cache probe.
     pub event_name: Option<&'a str>,
-    /// `true` when this is the first time this schema has been seen
-    /// (cache miss).  Exporters can use this to register the
-    /// `EventFormat` for a unique ID without checking every event.
-    pub is_new_schema: bool,
+    /// Monotonic identifier for this event's schema.
+    ///
+    /// Each distinct `(schema_tl_bytes, pointer_width)` pair gets a unique
+    /// `SchemaId` on first insertion into the cache.  The same ID is
+    /// returned on every subsequent cache hit.  Exporters can use this as
+    /// a cheap lookup key to avoid re-registering the `EventFormat`:
+    ///
+    /// ```ignore
+    /// let event = decoder.decode(record)?;
+    /// let id = *my_map.entry(event.schema_id)
+    ///     .or_insert_with(|| exporter.register(event.event_data.format()));
+    /// ```
+    pub schema_id: SchemaId,
 }
+
+/// Opaque monotonic identifier for a cached TDH schema.
+///
+/// Assigned once on cache insertion and returned with every decoded event
+/// that matches the same schema.  Suitable as a `HashMap` key for
+/// exporter-side format registration.
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Hash)]
+pub struct SchemaId(u64);
 
 /// Runtime decoder for TraceLogging / TraceLoggingDynamic ETW events.
 ///
@@ -198,6 +216,8 @@ pub struct TdhDecoder {
     cache: SchemaCache,
     /// Reusable aligned buffer for `TdhGetEventInformation` results.
     tei_buf: AlignedTeiBuf,
+    /// Monotonic counter for assigning unique `SchemaId`s.
+    next_schema_id: u64,
 }
 
 impl TdhDecoder {
@@ -206,16 +226,16 @@ impl TdhDecoder {
         Self {
             cache: SchemaCache::new(),
             tei_buf: AlignedTeiBuf::new(),
+            next_schema_id: 0,
         }
     }
 
     /// Decodes an `EVENT_RECORD` into a [`TdhDecodedEvent`].
     ///
     /// The returned [`TdhDecodedEvent`] contains the decoded
-    /// [`EventData`] together with `is_new_schema`, which is `true`
-    /// when this schema was seen for the first time (cache miss).
-    /// Exporters can use this flag to register the [`EventFormat`]
-    /// for a unique ID without checking on every event.
+    /// [`EventData`], the TraceLogging `event_name`, and a monotonic
+    /// [`SchemaId`] that exporters can use as a cheap lookup key for
+    /// format registration.
     pub fn decode<'a>(
         &'a mut self,
         record: &'a EVENT_RECORD,
@@ -223,20 +243,21 @@ impl TdhDecoder {
         let is_32bit = (record.EventHeader.Flags & EVENT_HEADER_FLAG_32_BIT_HEADER) != 0;
         let schema_tl_bytes = find_schema_tl(record)?;
 
-        let is_new_schema = if self.cache.get(schema_tl_bytes, is_32bit).is_none() {
+        if self.cache.get(schema_tl_bytes, is_32bit).is_none() {
             call_tdh_get_event_information(record, &mut self.tei_buf)?;
-            let schema = build_cached_schema(self.tei_buf.as_bytes(), is_32bit)?;
+            let mut schema = build_cached_schema(self.tei_buf.as_bytes(), is_32bit)?;
+            let id = self.next_schema_id;
+            self.next_schema_id = id.wrapping_add(1);
+            schema.schema_id = SchemaId(id);
             debug!(
                 event_name = %schema.event_name,
                 field_count = schema.format.fields().len(),
+                schema_id = id,
                 is_32bit,
                 "TDH schema cache miss — new schema cached"
             );
             self.cache.insert(schema_tl_bytes.to_vec(), is_32bit, schema);
-            true
-        } else {
-            false
-        };
+        }
 
         let schema = self.cache.get(schema_tl_bytes, is_32bit)
             .expect("just inserted");
@@ -256,7 +277,7 @@ impl TdhDecoder {
         Ok(TdhDecodedEvent {
             event_data: EventData::new(user_data, user_data, &schema.format),
             event_name,
-            is_new_schema,
+            schema_id: schema.schema_id,
         })
     }
 }
@@ -287,6 +308,7 @@ fn build_cached_schema(tei_buf: &[u8], is_32bit: bool) -> Result<CachedSchema, T
         return Ok(CachedSchema {
             event_name,
             format: EventFormat::new(),
+            schema_id: SchemaId(0), // assigned by caller
         });
     }
 
@@ -323,6 +345,7 @@ fn build_cached_schema(tei_buf: &[u8], is_32bit: bool) -> Result<CachedSchema, T
     Ok(CachedSchema {
         event_name,
         format,
+        schema_id: SchemaId(0), // assigned by caller
     })
 }
 
@@ -785,7 +808,7 @@ mod tests {
 
         let mut decoder = TdhDecoder::new();
         let result = decoder.decode(&record).expect("decode should succeed");
-        assert!(result.is_new_schema);
+        let _schema_id = result.schema_id;
 
         let fields = result.event_data.format().fields();
         assert_eq!(fields.len(), 1);
@@ -941,14 +964,14 @@ mod tests {
 
         let mut decoder = TdhDecoder::new();
         let r1 = decoder.decode(&record_1).expect("first decode");
-        assert!(r1.is_new_schema);
+        let id1 = r1.schema_id;
         assert_eq!(r1.event_data.format().fields()[0].size, 4);
 
         let user_data_2 = 222u32.to_le_bytes();
         let mut ext_items_2: [EVENT_HEADER_EXTENDED_DATA_ITEM; 2] = unsafe { std::mem::zeroed() };
         let record_2 = build_test_record(prov, evt, &mut ext_items_2, &user_data_2);
         let r2 = decoder.decode(&record_2).expect("second decode (cached)");
-        assert!(!r2.is_new_schema);
+        assert_eq!(r2.schema_id, id1, "cache hit should return same SchemaId");
         assert_eq!(r2.event_data.format().fields()[0].size, 4);
     }
 
@@ -986,7 +1009,7 @@ mod tests {
 
         let mut decoder = TdhDecoder::new();
         let r64 = decoder.decode(&record_64).expect("64-bit decode");
-        assert!(r64.is_new_schema, "64-bit should be a cache miss");
+        let id_64 = r64.schema_id;
 
         // 32-bit decode (set EVENT_HEADER_FLAG_32_BIT_HEADER)
         let mut ext_items_32: [EVENT_HEADER_EXTENDED_DATA_ITEM; 2] = unsafe { std::mem::zeroed() };
@@ -994,19 +1017,20 @@ mod tests {
         record_32.EventHeader.Flags |= EVENT_HEADER_FLAG_32_BIT_HEADER;
 
         let r32 = decoder.decode(&record_32).expect("32-bit decode");
-        assert!(r32.is_new_schema, "32-bit should also be a cache miss (separate cache)");
+        let id_32 = r32.schema_id;
+        assert_ne!(id_64, id_32, "32-bit and 64-bit should get different SchemaIds");
 
-        // Second 64-bit decode should hit the 64-bit cache
+        // Second 64-bit decode should return the same SchemaId
         let mut ext_items_64b: [EVENT_HEADER_EXTENDED_DATA_ITEM; 2] = unsafe { std::mem::zeroed() };
         let record_64b = build_test_record(prov, evt, &mut ext_items_64b, &user_data);
         let r64b = decoder.decode(&record_64b).expect("64-bit cache hit");
-        assert!(!r64b.is_new_schema, "64-bit should be a cache hit");
+        assert_eq!(r64b.schema_id, id_64, "64-bit cache hit should return same ID");
 
-        // Second 32-bit decode should hit the 32-bit cache
+        // Second 32-bit decode should return the same SchemaId
         let mut ext_items_32b: [EVENT_HEADER_EXTENDED_DATA_ITEM; 2] = unsafe { std::mem::zeroed() };
         let mut record_32b = build_test_record(prov, evt, &mut ext_items_32b, &user_data);
         record_32b.EventHeader.Flags |= EVENT_HEADER_FLAG_32_BIT_HEADER;
         let r32b = decoder.decode(&record_32b).expect("32-bit cache hit");
-        assert!(!r32b.is_new_schema, "32-bit should be a cache hit");
+        assert_eq!(r32b.schema_id, id_32, "32-bit cache hit should return same ID");
     }
 }
--- a/one_collect/src/etw/tdh.rs
+++ b/one_collect/src/etw/tdh.rs
@@ -174,9 +174,16 @@ impl SchemaCache {
 /// whether this was the first time the schema was seen (cache miss).
 /// Callers can use `is_new_schema` to register the [`EventFormat`]
 /// with an exporter exactly once.
-pub struct DecodeResult<'a> {
+#[non_exhaustive]
+pub struct TdhDecodedEvent<'a> {
     /// The decoded event data.
     pub event_data: EventData<'a>,
+    /// The TraceLogging event name, or `None` if the schema has no name.
+    ///
+    /// This is the primary identity for TraceLogging events (as opposed
+    /// to the event ID, which is often 0).  Consumers can use this for
+    /// OTEL log record naming without a second cache probe.
+    pub event_name: Option<&'a str>,
     /// `true` when this is the first time this schema has been seen
     /// (cache miss).  Exporters can use this to register the
     /// `EventFormat` for a unique ID without checking every event.
@@ -202,22 +209,9 @@ impl TdhDecoder {
         }
     }
 
-    /// Returns the cached event name for the given event's schema, or
-    /// `None` if the schema has not been seen yet or has no name.
-    pub fn event_name(&self, record: &EVENT_RECORD) -> Option<&str> {
-        let is_32bit = (record.EventHeader.Flags & EVENT_HEADER_FLAG_32_BIT_HEADER) != 0;
-        let schema_tl_bytes = find_schema_tl(record).ok()?;
-        let schema = self.cache.get(schema_tl_bytes, is_32bit)?;
-        if schema.event_name.is_empty() {
-            None
-        } else {
-            Some(&schema.event_name)
-        }
-    }
-
-    /// Decodes an `EVENT_RECORD` into a [`DecodeResult`].
+    /// Decodes an `EVENT_RECORD` into a [`TdhDecodedEvent`].
     ///
-    /// The returned [`DecodeResult`] contains the decoded
+    /// The returned [`TdhDecodedEvent`] contains the decoded
     /// [`EventData`] together with `is_new_schema`, which is `true`
     /// when this schema was seen for the first time (cache miss).
     /// Exporters can use this flag to register the [`EventFormat`]
@@ -225,7 +219,7 @@ impl TdhDecoder {
     pub fn decode<'a>(
         &'a mut self,
         record: &'a EVENT_RECORD,
-    ) -> Result<DecodeResult<'a>, TdhDecodeError> {
+    ) -> Result<TdhDecodedEvent<'a>, TdhDecodeError> {
         let is_32bit = (record.EventHeader.Flags & EVENT_HEADER_FLAG_32_BIT_HEADER) != 0;
         let schema_tl_bytes = find_schema_tl(record)?;
 
@@ -253,8 +247,15 @@ impl TdhDecoder {
             field_count = schema.format.fields().len(),
             "TDH decode — user_data"
         );
-        Ok(DecodeResult {
+        let event_name = if schema.event_name.is_empty() {
+            None
+        } else {
+            Some(schema.event_name.as_str())
+        };
+
+        Ok(TdhDecodedEvent {
             event_data: EventData::new(user_data, user_data, &schema.format),
+            event_name,
             is_new_schema,
         })
     }
@@ -348,7 +349,8 @@ fn walk_properties(
         }
 
         let prop = &properties[i];
-        let name = read_property_name(tei_buf, prop);
+        let raw_name = read_property_name(tei_buf, prop);
+        let name = if raw_name.is_empty() { std::format!("field{i}") } else { raw_name };
         let qualified_name = if prefix.is_empty() {
             name
         } else {
@@ -599,7 +601,7 @@ fn read_utf16_at(buf: &[u8], byte_offset: usize) -> String {
 /// This is the single source of truth for the in-type → field-info
 /// mapping, used by both `walk_properties` (schema construction) and
 /// `intype_to_type_name` (explicit-length fallback).
-fn intype_to_field_info(in_type: i32, is_32bit: bool) -> (&'static str, LocationType, usize) {
+const fn intype_to_field_info(in_type: i32, is_32bit: bool) -> (&'static str, LocationType, usize) {
     match in_type {
         // Fixed-size scalars
         TDH_INTYPE_INT8                          => ("s8",   LocationType::Static, 1),
@@ -649,7 +651,7 @@ fn intype_to_field_info(in_type: i32, is_32bit: bool) -> (&'static str, Location
 /// Returns the type_name string for a TDH_INTYPE.
 ///
 /// Delegates to [`intype_to_field_info`] and returns just the name.
-fn intype_to_type_name(in_type: i32) -> &'static str {
+const fn intype_to_type_name(in_type: i32) -> &'static str {
     intype_to_field_info(in_type, false).0
 }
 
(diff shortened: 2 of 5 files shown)


one_collect/src/etw/tdh.rs
```diff
@@ -0,0 +1,995 @@
...
+        };
+
+        let flags = prop.Flags;
+
+        // ── Struct property ─────────────────────────────────────────
+        if (flags & PROPERTY_STRUCT) != 0 {
+            let struct_info = unsafe { prop.Anonymous1.structType };
+            let start = struct_info.StructStartIndex as usize;
+            let count = struct_info.NumOfStructMembers as usize;
+            walk_properties_to_cache(
```
beaubelgrave: Lets have some depth parameter so we can prevent bad payloads of TDH from consuming too much memory or cycles. A reasonable limit of nesting depth should cause a warning and truncation, but not stop.

one_collect/src/etw/tdh.rs
```diff
@@ -0,0 +1,995 @@
...
+        unsafe {
+            std::slice::from_raw_parts(
+                record.UserData as *const u8,
+                record.UserDataLength as usize,
+            )
+        }
+    }
+}
+
+fn find_schema_tl<'a>(record: &'a EVENT_RECORD) -> Result<&'a [u8], TdhDecodeError> {
```
beaubelgrave, line 503: See etw/mod.rs under AncilaryData, there are functions to find extensions by type already. Lets move searching for extended items to a central place (like an impl on the EVENT_RECORD itself).

one_collect/src/etw/tdh.rs
```diff
@@ -0,0 +1,995 @@
...
+        warn!(win32_error = status, "TdhGetEventInformation sizing call failed");
+        return Err(TdhDecodeError::Win32(status));
+    }
+    if buffer_size == 0 {
+        warn!("TdhGetEventInformation returned zero buffer size");
+        return Err(TdhDecodeError::Malformed("TDH returned zero buffer size"));
+    }
+    // Allocate with u64 alignment (8 bytes), rounding up.
+    let u64_count = (buffer_size as usize + 7) / 8;
+    let mut storage: Vec<u64> = vec![0u64; u64_count];
```
beaubelgrave: We hide this vec allocation, which prevents sharing a re-usable buffer from callbacks and others. Can we move to a pattern where the AlignedTeiBuf is allocated once and re-used on a per-session or event basis to limit alloc/free patterns?

one_collect/src/etw/tdh.rs
```diff
@@ -0,0 +1,995 @@
...
+    fn scan_counted_truncated() {
+        // len=10 but only 4 bytes available after prefix
+        let data: &[u8] = &[0x0A, 0x00, 0x01, 0x02, 0x03, 0x04];
+        assert_eq!(scan_counted_string(data), 6); // clamped to data.len()
+    }
+
+    // ── format with counted string ──────────────────────────────────
+
+    #[test]
+    fn format_with_counted_string() {
```
beaubelgrave: Can we get TDH / TL Schema bytes so we can have unit tests for the full end-to-end including the underlying TdhAPI calls (when running on Windows)?

beaubelgrave (changes requested):
It's looking good, some comments I'd like addressed. Mainly re-using memory for the AlignedTeiBuf seems like a useful thing to do by scanning where it is used (simply to walk and copy data out it seems). But let me know if that is prevented by something.

one_collect/src/etw/tdh.rs
```diff
@@ -0,0 +1,995 @@
...
+const EVENT_HEADER_FLAG_32_BIT_HEADER: u16 = 0x0020;
+
+/// Property flag constants from `EVENT_PROPERTY_INFO::Flags` (i32 in windows-sys).
+const PROPERTY_STRUCT: i32      = 0x1;
+const PROPERTY_HAS_LENGTH: i32  = 0x4;
+const PROPERTY_HAS_COUNT: i32   = 0x8;
+/// `PROPERTY_PARAM_LENGTH` — when set together with `PROPERTY_HAS_LENGTH`,
+/// `Anonymous3.length` is an *index* into the property array (not a literal
+/// byte count).  We must not treat it as a fixed size in that case.
+const PROPERTY_PARAM_LENGTH: i32 = 0x10;
```
utpilla: These four constants don't line up with `PROPERTY_FLAGS` in the [Windows SDK](https://learn.microsoft.com/en-us/windows/win32/api/tdh/ne-tdh-property_flags).

Could we import these constants from [`windows-sys`](https://docs.rs/windows-sys/0.61.2/windows_sys/Win32/System/Diagnostics/Etw/index.html) crate instead of hand-rolling them? That would lead to lesser correctness issues. For example, `PropertyParamLength` should be `2` but it's set to `10` here.

one_collect/src/etw/tdh.rs
```diff
@@ -0,0 +1,995 @@
...
+
+/// Extended-data item type for TraceLogging schema metadata.
+const EVENT_HEADER_EXT_TYPE_EVENT_SCHEMA_TL: u16 = 11;
```
utpilla: This could also be used from `windows-sys` crate.

one_collect/src/etw/tdh.rs
```diff
@@ -0,0 +1,995 @@
...
+
+// ── TdhDecoder ──────────────────────────────────────────────────────
+
+/// Runtime decoder for TraceLogging / TraceLoggingDynamic ETW events.
+pub struct TdhDecoder {
+    cache: SchemaLayoutCache,
+    /// Reusable `EventFormat` buffer to avoid per-event allocation
+    /// when the schema has variable-length fields.
+    format_buf: EventFormat,
+}
```
utpilla, line 221: The decoder doesn't follow the framework's existing conventions for variable-length fields. The framework already has a well-established way of describing events whose layout depends on payload contents (length-prefixed arrays, null-terminated strings, etc.). The schema declares which fields are variable and what type they are; the framework resolves their actual byte positions lazily, at the moment a consumer asks for them. This PR opts out of that convention and implements a parallel mechanism which comes at a cost.

**1. The cache caches the wrong thing.**

Conceptually, what changes across events of the same schema is only the payload bytes. What doesn't change is the schema (field names, types, structural shape). A useful cache should store the part that doesn't change.

This PR caches schema fragments (`Vec<CachedProperty>`) but rebuilds the consumer-facing artifact (`EventFormat`) on every event, because the per-event model bakes payload-specific sizes into each field. Net effect: the cache saves the one expensive thing (`TdhGetEventInformation`), but every hit still pays for a full schema walk and a fresh allocation.

If the `EventFormat` itself were schema-stable (not payload-dependent), it could be cached in full, and a cache hit collapses to a hashmap probe plus an `EventData::new` (i.e. effectively free). That's the shape every other event source in one_collect already has.

**2. Eager work that should be lazy.**

`build_event_format_with_data` resolves every variable field's byte length up-front on every event, even fields the consumer never reads. That's eager work in a hot path. Consumers typically project events down to a handful of fields; everything else is wasted scanning.

The framework's convention is the opposite: describe variability at schema time, resolve byte positions only when someone calls for the field. Skipped fields cost nothing. This could matter a lot at `TraceLogging` event rates.

**3. Re-implementing existing infrastructure.**

`scan_ansi_string`, `scan_utf16_string`, `scan_counted_string` are reimplementations of payload-walking logic that already lives in event/mod.rs ([get_field_data_closure](https://github.com/microsoft/one-collect/blob/f655a30aec1d15467501ae9f735defb2d6b86581/one_collect/src/event/mod.rs#L592-L660) and the skip-chain machinery). It's best to avoid two copies of the same logic. The cleanest version of this PR should not have this scanning logic at all; that work belongs to the framework, where it's already battle-tested by every other event source.

**4. Compatibility with the rest of the system.**

Scripting, filters, the universal exporter — all of them work against `EventFormat` and reuse `try_get_field_data_closure` results across events of the same schema. They assume the format is a property of the schema, not of any particular event. A per-event mutating format breaks that assumption silently: a closure (or compiled script) prepared against one event reads the wrong bytes from the next. By adopting the standard convention, TDH-decoded events compose with everything else for free.

**5. Less code, fewer concepts.**

Following the convention removes `PropertyKind`, `CachedProperty`, the schema-fragment-vs-format split, `build_event_format_with_data`, `format_buf`, the three scan helpers, and the fast-path/slow-path branching. It would simplify things to "walk the TDH property list once, emit `EventFields`, store the resulting `EventFormat`".

**What "following the convention" means concretely**

When walking the TDH properties at cache-insert time:
- Fixed scalars → `Static` with their byte width and running absolute offset.
- Null-terminated strings → `StaticString` / `StaticUTF16String` with `size = 0`.
- Length-prefixed strings → `StaticLenPrefixArray` with `type_name = "u8"` and `size = 0`.
- After the first variable field, subsequent fields' literal offset is irrelevant (set to `0` by convention); the framework's skip chain resolves them at read time.

The reason `size = 0` is load-bearing rather than cosmetic: the framework's skip walker (`try_get_field_data_closure` in `event/mod.rs`) only treats a field as variable when `size == 0`. Anything non-zero is taken to mean "the next field's offset already accounts for these bytes". The two encodings aren't interchangeable.

**Summary of what we gain by adopting the convention**

- Cache becomes effective (stores `EventFormat`, not fragments) → hot path is a hashmap probe.
   ```diff
   - cache: HashMap<Vec<u8>, CachedSchema>   // keyed by schema-TL bytes
   + cache: HashMap<Vec<u8>, EventFormat>    // keyed by schema-TL bytes (+ pointer width)
   ```
- Work is lazy → only fields the consumer reads cost anything.
- No re-implementation of scan logic → one source of truth in `event/mod.rs`.
- TDH-decoded events compose with scripting / filters / exporters with no special-casing.
- Significantly less code, fewer concepts to maintain.

one_collect/src/etw/mod.rs
```diff
@@ -83,6 +87,13 @@ impl AncillaryData {
...
     }
 
+    /// Exposes the raw EVENT_RECORD pointer for TDH decoding.
+    /// Returns None if no event is currently being processed.
+    pub fn raw_event_record(&self) -> Option<*const EVENT_RECORD> {
+        self.event
+    }
+
+
     pub fn time(&self) -> u64 {
```
utpilla: Two suggested changes here:
1. Let's return a borrow instead of a raw pointer. The pointer is only valid while the callback runs which is precisely what `&self` already encodes. This removes an `unsafe { &* }` from every call site, and the safety contract gets pushed down to the one place that knows the pointer is valid.
2. Removed the extra blank line before the `time` method.

```suggestion
    pub fn record(&self) -> Option<&EVENT_RECORD> {
        self.event.map(|p| unsafe { &*p })
    }

pub fn time(&self) -> u64 {
```

one_collect/src/etw/tdh.rs
```diff
@@ -0,0 +1,995 @@
...
+    if property_count == 0 {
+        return Ok(CachedSchema {
+            event_name,
+            properties: Vec::new(),
+            fixed_format: Some(EventFormat::new()),
+        });
+    }
+
+    let props_offset = std::mem::size_of::<TRACE_EVENT_INFO>();
+    let props_end = props_offset + property_count * std::mem::size_of::<EVENT_PROPERTY_INFO>();
```
utpilla: It might be unlikely, but it would be good to use `checked_add` / `checked_mul` here and return `TdhDecodeError::Malformed(...)` on overflow.

one_collect/src/etw/tdh.rs
```diff
@@ -0,0 +1,995 @@
...
+/// based decode (#4).
+fn read_property_name(tei_buf: &[u8], prop: &EVENT_PROPERTY_INFO) -> String {
+    read_utf16_at(tei_buf, prop.NameOffset as usize)
+}
+
+/// Reads a null-terminated UTF-16LE string from `buf` at `byte_offset`.
+///
+/// Returns an empty string if the offset is out of bounds.  Uses
+/// `String::from_utf16_lossy` but feeds it from an iterator to avoid
+/// an intermediate `Vec<u16>` allocation for small names (#4).
```
utpilla: This method does allocate though:

```rust
// Collect into a small vec — property names are typically short.
let u16s: Vec<u16> = u16_iter.collect();
```

one_collect/src/etw/tdh.rs
```diff
@@ -0,0 +1,995 @@
...
+            format_buf: EventFormat::new(),
+        }
+    }
+
+    /// Returns the cached event name for the given event's schema, or
+    /// `None` if the schema has not been seen yet or has no name.
+    ///
+    /// This is the TraceLogging event name — the primary identity for
+    /// TraceLogging events (as opposed to the event ID, which is often 0).
+    pub fn event_name(&self, record: &EVENT_RECORD) -> Option<&str> {
```
utpilla: Is this method needed?

one_collect/src/etw/tdh.rs
```diff
@@ -0,0 +1,995 @@
...
+    let status = unsafe {
+        TdhGetEventInformation(
+            record as *const EVENT_RECORD, 0u32,
+            core::ptr::null(), core::ptr::null_mut(), &mut buffer_size,
+        )
+    };
+    if status != ERROR_INSUFFICIENT_BUFFER {
+        warn!(win32_error = status, "TdhGetEventInformation sizing call failed");
+        return Err(TdhDecodeError::Win32(status));
+    }
```
utpilla, line 562: Minor: We don't expect this method to be called for non-Tracelogging events but if they are we can return `TdhError::NotFound`.

```suggestion
    if status == ERROR_NOT_FOUND {
        return Err(TdhDecodeError::NotFound);
    }
    
    if status != ERROR_INSUFFICIENT_BUFFER {
        warn!(win32_error = status, "TdhGetEventInformation sizing call failed");
        return Err(TdhDecodeError::Win32(status));
    }
```

one_collect/src/etw/tdh.rs
```diff
@@ -0,0 +1,995 @@
...
+    if tei_buf.len() < EVENT_NAME_OFFSET_POS + 4 {
+        return String::new();
+    }
+    let name_offset_bytes: [u8; 4] = tei_buf[EVENT_NAME_OFFSET_POS..EVENT_NAME_OFFSET_POS + 4]
+        .try_into()
+        .unwrap();
+    let name_offset = u32::from_le_bytes(name_offset_bytes) as usize;
+
+    // Fallback: try the struct field directly if available.
+    let _ = tei; // suppress unused warning; we use the raw buffer above.
```
utpilla: The hardcoded offset `92` doesn't help: union access still requires unsafe either way (raw-byte access just swaps type-checked for offset-checked), and the offset is the running sum of every preceding field's size. Any future windows-sys field reorder would shift it silently with no compile-time signal. The `tei: &TRACE_EVENT_INFO` parameter is also unused (`let _ = tei;`), pointing at the same confusion.

In `windows-sys 0.59` both arms of `TRACE_EVENT_INFO_0` are `u32` at the same offset, so either side always yields a valid bit pattern. We have a similar situation already handled correctly in `abi.rs` for `BufferContext.Anonymous.ProcessorIndex`. We could do something like:

```rust
fn read_event_name(tei_buf: &[u8], tei: &TRACE_EVENT_INFO) -> String {
    // SAFETY: Both arms of TRACE_EVENT_INFO_0 are `u32` at the same offset,
    // so either arm always yields a valid bit pattern.
    let name_offset = unsafe { tei.Anonymous1.EventNameOffset } as usize;
    read_utf16_at(tei_buf, name_offset)
}
```

We could also add a `const` assertion to protect from future drift:

```rust
const _: () = assert!(
    std::mem::size_of::<windows_sys::Win32::System::Diagnostics::Etw::TRACE_EVENT_INFO_0>() == 4,
    "TRACE_EVENT_INFO_0 union must remain 4 bytes",
);
```

one_collect/src/etw/mod.rs
```diff
@@ -15,11 +15,15 @@ use crate::Guid;
 #[allow(dead_code)]
 mod abi;
 mod events;
+pub mod tdh;
+
+/// Re-export `EVENT_RECORD` so consumers of `TdhDecoder` can name the
+/// type without reaching into `abi` (which is `pub(crate)`).
+pub use abi::EVENT_RECORD;
```
utpilla: I suggest dropping this re-export. Consumers already get `windows-sys` transitively through one_collect (it's in the `Cargo.toml` as a direct dep with `Win32_System_Diagnostics_Etw`). It would be cleaner to have `TdhDecoder::decode` use `&windows_sys::…::EVENT_RECORD`.

one_collect/src/etw/tdh.rs
swashtek: `event_name()` enables the consuming ETW receiver to use TraceLogging event names for OTEL log record event names. The current `"etw.0"` naming for all TL events is a known limitation that this API would fix. Will address it in follow up PR.

one_collect/src/etw/tdh.rs
swashtek: Yes, that is correct, let me remove misleading comment.

one_collect/src/etw/tdh.rs
```diff
@@ -0,0 +1,919 @@
...
+    /// Decodes an `EVENT_RECORD` into an [`EventData`].
+    ///
+    /// The returned `EventData` references the cached `EventFormat`
+    /// (schema-stable).  Consumers use the framework's standard
+    /// `try_get_field_data_closure` to resolve individual field values
+    /// lazily against the event's `user_data` payload.
+    pub fn decode<'a>(
+        &'a mut self,
+        record: &'a EVENT_RECORD,
+    ) -> Result<EventData<'a>, TdhDecodeError> {
```
beaubelgrave: In the exporting case, we will want to know if we've seen this event format before. We need a way to identify if it's a new item within the cache or existing.

In exporting cases, we would pass the EventFormat to the exporter to get a unique ID to label events with. We wouldn't want to do this for every event, only unique EventFormats.

swashtek: Makes sense, let me add a boolean `is_new_schema` and thanks for explaining the use case.

one_collect/src/etw/tdh.rs
```diff
@@ -458,15 +462,40 @@ fn walk_properties(
...
-            TDH_INTYPE_UNICODESTRING |
-            TDH_INTYPE_NONNULLTERMINATEDSTRING       => ("wstring", LocationType::StaticUTF16String, 0),
-
-            // Variable-length: counted/length-prefixed
-            TDH_INTYPE_COUNTEDSTRING |
-            TDH_INTYPE_REVERSEDCOUNTEDSTRING         => ("counted_string", LocationType::StaticLenPrefixArray, 0),
+            TDH_INTYPE_ANSISTRING                    => ("string",  LocationType::StaticString, 0),
+            TDH_INTYPE_UNICODESTRING                 => ("wstring", LocationType::StaticUTF16String, 0),
+
+            // Extended string types (300+ range).
```
swashtek: @Utpilla There is a mismatch between windows-sys and TDH runtime api, so handling 300+ entries differently for old and new SDKs. Wdyt ?

utpilla: When you say "old SDK / new SDK", which Windows SDK version made which enum change, and on which Windows runtime did you observe the mismatch?

Asking because the inline comment commits to a specific historical claim:

```rust
// Old SDK:  300=Counted, 301=ReversedCounted,
//           302=NonNullTerm, 303=NonNullTermAnsi
// New SDK:  300=Counted, 301=CountedAnsi,
//           302=ReversedCounted, 303=ReversedCountedAnsi,
//           304=NonNullTerm, 305=NonNullTermAnsi
```

Two things don't line up for me:

1. The official enum reference ([`_TDH_IN_TYPE` on Microsoft Learn](https://learn.microsoft.com/en-us/windows/win32/api/tdh/ne-tdh-_tdh_in_type)) lists the names in this declaration order: …`COUNTEDSTRING`, `COUNTEDANSISTRING`, `REVERSEDCOUNTEDSTRING`, `REVERSEDCOUNTEDANSISTRING`, NONNULLTERMINATEDSTRING, NONNULLTERMINATEDANSISTRING…. So `CountedAnsi` comes immediately after `Counted`. The "Old SDK" line putting `ReversedCounted` in that slot doesn't match the documented order.
2. windows-sys 0.59 has the explicit numeric values 300–305 in that same order, matching the doc.

Could you share where the "Old SDK" numbering comes from? Is it a Windows SDK version, an older `tdh.h`, a runtime observation? I want to make sure we're guarding against a real mismatch before we land a literal-integer table that bypasses the typed constants; if there isn't one, I'd lean toward reverting to the typed `TDH_INTYPE_*` constants from `windows-sys`.

one_collect/src/etw/tdh.rs
```diff
@@ -0,0 +1,995 @@
...
+
+/// `EVENT_HEADER_FLAG_32_BIT_HEADER` from the Windows SDK.
+const EVENT_HEADER_FLAG_32_BIT_HEADER: u16 = 0x0020;
+
+// Aliases for PROPERTY_FLAGS constants (i32 in windows-sys) to keep
+// call-site flag checks concise.
+const PROPERTY_STRUCT: i32             = PropertyStruct;
+const PROPERTY_PARAM_LENGTH: i32       = PropertyParamLength;
+const PROPERTY_PARAM_COUNT: i32        = PropertyParamCount;
+const PROPERTY_PARAM_FIXED_LENGTH: i32 = PropertyParamFixedLength;
```
utpilla: This is unused. Do we need it?

one_collect/src/etw/tdh.rs
```diff
@@ -0,0 +1,995 @@
...
+                    max = MAX_STRUCT_DEPTH,
+                    "struct nesting depth exceeded — truncating sub-fields"
+                );
+                // Emit a placeholder that consumes remaining bytes.
+                let offset = if *seen_variable { 0 } else { *running_offset };
+                format.add_field(EventField::new(
+                    qualified_name, "unsupported".to_string(),
+                    LocationType::Static, offset, 0,
+                ));
+                *seen_variable = true;
```
utpilla: When the depth limit fires, we emit a placeholder `(LocationType::Static, size = 0, type_name = "unsupported")`, flip `seen_variable = true`, keep walking the rest of the parent's properties, and cache the resulting `EventFormat`. That cache entry is then permanently broken for every event of this schema:

- The placeholder is `(Static, size = 0)`, which `try_get_field_data_closure` doesn't treat as a variable field. It isn't added to the skip chain. Reading the placeholder itself gets an empty slice, which is at least visible.
- More importantly, every field declared after the placeholder becomes unreadable too, because the skip-chain walker has no way to advance past a `(Static, size = 0)` entry. `try_get_field_data_closure` returns `None` for all of them.
- The `warn!` fires exactly once (at cache-insert time). Every subsequent event of this schema silently loses data with no further signal.

Suggest returning `Err(TdhDecodeError::Malformed("struct nesting too deep"))` instead. That variant already exists for this kind of case; the caller gets a visible, named failure on every event, and no half-correct format gets baked into the cache. Yes, that drops the whole event rather than the over-deep subtree but 8-level-deep nesting is rare enough that "fail loudly and let the caller decide" is the lower-cost failure mode versus "silently drop the tail of every event forever".

one_collect/src/etw/tdh.rs
utpilla: @beaubelgrave @swashtek The use case makes sense, but the way we're surfacing it carries a subtle commitment.

`is_new_schema = true` means "cache miss on this call". Today that's equivalent to "first time we've ever seen this schema" because nothing ever leaves the cache. The two stop being the same the moment we add any kind of bounded / LRU eviction, which seems plausible for long-running collectors against `TraceLoggingDynamic` providers:

- Schema X evicted → arrives again → `is_new_schema = true` even though the consumer has seen and registered it before.
- The flag stops being a reliable "register this with the exporter now" signal. Consumers would have to keep their own `HashMap<SchemaKey, ExporterId>` and dedup against it anyway (the exact per-event work this flag was meant to save).

Worth thinking about whether we want to surface this signal in the public API at all, given the constraint it locks in.

one_collect/src/event/mod.rs
```diff
@@ -1156,10 +1166,11 @@ impl EventFormat {
                 }
 
                 match type_name {
-                    "u8" | "s8" | "char" => Some(1),
-                    "u16" | "s16" | "short" => Some(2),
-                    "u32" | "s32" | "int" => Some(4),
-                    "u64" | "s64" | "long" => Some(8),
+                    "u8" | "s8" | "char" | "counted_string"
+                    | "string" | "wstring"                   => Some(1),
```
utpilla: The TDH decoder emits counted strings as `(StaticLenPrefixArray, "wstring", size = 0)` so the framework's existing write-closure renders them as text. That's what required `try_get_element_size` to learn about `"wstring"/"string"/"counted_string" and return Some(1)`.

One ask: please add a one-liner comment at the new `try_get_element_size` arm explaining that `Some(1)` is intentional. The value is a byte count from the wire format, not a code-unit count. Without that note, `Some(1)` looks wrong to anyone who hasn't traced the read path (UTF-16 code units are 2 bytes wide, so the natural "fix" is `Some(2)`, which would silently halve every counted-string read).

one_collect/src/event/mod.rs
```diff
@@ -566,15 +566,24 @@ impl EventFormat {
...
+                /* When size is 0, default to byte-counted (element_size=1).
+                 * This handles counted string fields where the u16 prefix
+                 * is a byte count. */
+                let element_size = if size == 0 { 1 } else { size };
+                let bytes = len * element_size;
+
+                if 2 + bytes > slice.len() {
+                    /* Out of bounds */
+                    return EMPTY;
+                }
```
utpilla, line 586: nit: We could write some unit tests independent of TDH to verify these fixes.

one_collect/src/etw/tdh.rs
```diff
@@ -0,0 +1,995 @@
...
+                LocationType::Static, offset, len,
+            ));
+            if !*seen_variable {
+                *running_offset += len;
+            }
+            continue;
+        }
+
+        // Map TDH in-type to the framework's LocationType + size.
+        let (type_name, loc, size) = match in_type {
```
utpilla: There are two copies of the mapping. One is here and the other is in `intype_to_type_name` method.

We should be able to return `(name, loc, size)` from one mapping function and call it from both sites, or derive the name from the kind/location to avoid the duplication.

one_collect/src/etw/tdh.rs
swashtek: You are correct here, the actual bug is different,

**Summary of the root cause evolution:**

1. **Original code** (commit `89aaa5f`): Used typed `TDH_INTYPE_*` constants from `windows-sys` correctly, but treated both `COUNTEDSTRING` (300, UTF-16) and `COUNTEDANSISTRING` (301, UTF-8) as the same type `"counted_string"`. The consumer then decoded all counted strings as UTF-16.

2. **My incorrect "fix"** (commit `0b73712`): I replaced the typed constants with integer literals `300..=305` and added a fabricated "Old SDK vs New SDK" comment claiming the enum values shifted between SDK versions. This was wrong — the values have been stable. The real bug (treating 301/ANSI as UTF-16) was not addressed.

3. **Current fix**: 
   - Restored typed `TDH_INTYPE_*` constants from `windows-sys`
   - Split the type names: `"counted_wstring"` (300 = UTF-16) vs `"counted_string"` (301 = ANSI/UTF-8)
   - Consumer decodes `"counted_string"` as UTF-8 and `"counted_wstring"` as UTF-16
   - Removed all incorrect "old/new SDK" comments

The actual bug was always an **encoding mismatch**: the `tracelogging` crate's `str8()` writes UTF-8 counted strings (TDH in_type 301), but the decoder/consumer was treating all counted strings as UTF-16 — producing garbled CJK characters.

one_collect/src/etw/tdh.rs
swashtek: `PROPERTY_PARAM_FIXED_LENGTH` is unused, removing it.

one_collect/src/etw/tdh.rs
swashtek: Good point, refactoring to avoid duplication.

one_collect/src/etw/tdh.rs
beaubelgrave: Maybe the wording is_new_schema is the wrong one. Since exporting will have it's own lookups, the flag should somehow indicate it MIGHT be brand new. That way the lookup work can be avoided MOST of the time. An alternative is a solid ID, but those can be hard to guarantee uniqueness unless it's simply a monotonic ever increasing u64. That would give a global ID of the event without indicating if it's new or not, it just isn't guaranteed to be the same as a previous event (but it could be).

beaubelgrave: I think going with a monotonic ID for the schema of the event/payload is the way to go thinking about this more. That keeps the external lookup very small and accommodates for LRU removal later (at the cost of additional processing lookup when they do get removed, but nothing breaks).

one_collect/src/etw/tdh.rs
```diff
@@ -0,0 +1,953 @@
...
+
+// ── TdhDecoder ──────────────────────────────────────────────────────
+
+/// Result of a successful [`TdhDecoder::decode`] call.
+///
+/// Wraps the decoded [`EventData`] together with a flag indicating
+/// whether this was the first time the schema was seen (cache miss).
+/// Callers can use `is_new_schema` to register the [`EventFormat`]
+/// with an exporter exactly once.
+pub struct DecodeResult<'a> {
```
utpilla: Mark this struct non exhaustive to keep it open for addition of newer fields in future, if required.

```suggestion
#[non_exhaustive]
pub struct DecodeResult<'a> {
```

one_collect/src/etw/tdh.rs
utpilla: Makes sense. However, I would suggest we move `event_name` onto `DecodeResult` instead of offering it as a public method. Both `decode` and event_name` computation requires looking up the cache.

The current usage requires a second cache probe per event for data decode already had in hand:
```rust
let result = decoder.decode(record)?;
let name = decoder.event_name(record);   // re-walks ExtendedData, re-hashes the
                                          // schema-TL bytes, re-probes the cache
```

For a hot-path use case like OTEL log-record naming, this fires on every event.

Surfacing it as a field on `DecodeResult` is one field read off the `CachedSchema` reference decode already holds which is effectively free:

```rust
#[non_exhaustive]
pub struct DecodeResult<'a> {
    pub event_data: EventData<'a>,
    pub event_name: Option<&'a str>,
    pub is_new_schema: bool,
}
```

Caller code becomes:
```rust
let result = decoder.decode(record)?;
if let Some(name) = result.event_name { /* OTEL log record name */ }
```

Also removes the "must call `decode` first or you get `None`" precondition the current accessor quietly carries.

one_collect/src/etw/tdh.rs
```diff
@@ -0,0 +1,953 @@
...
+    ///
+    /// The returned [`DecodeResult`] contains the decoded
+    /// [`EventData`] together with `is_new_schema`, which is `true`
+    /// when this schema was seen for the first time (cache miss).
+    /// Exporters can use this flag to register the [`EventFormat`]
+    /// for a unique ID without checking on every event.
+    pub fn decode<'a>(
+        &'a mut self,
+        record: &'a EVENT_RECORD,
+    ) -> Result<DecodeResult<'a>, TdhDecodeError> {
```
utpilla: Could we rename `DecodeResult` to `TdhDecodedEvent`?

`Result` in Rust carries a specific meaning, and `Result<DecodeResult<'a>, TdhDecodeError>` doubles the word with two different intents in the same signature. A name that describes what the type is reads better at call sites:

```rust
let event = decoder.decode(record)?;
let name = event.event_name;
```

one_collect/src/etw/tdh.rs
utpilla: +1 to the monotonic-ID direction. That elegantly side-steps the contract problem with the boolean. The decoder stops trying to communicate "first time ever" semantics it can't actually guarantee under LRU, and instead just hands out cheap stable tokens that the exporter uses for its own bookkeeping.

Also, I'd suggest using an opaque (`pub struct SchemaId(u64)`) newtype so consumers can't do arithmetic on it.

Concretely:

```rust
#[non_exhaustive]
pub struct DecodedEvent<'a> {
    pub event_data: EventData<'a>,
    pub event_name: Option<&'a str>,
    pub schema_id: SchemaId,
}

#[derive(Copy, Clone, Debug, PartialEq, Eq, Hash)]
pub struct SchemaId(u64);
```

`decode` assigns the next monotonic `u64` on cache insert and hands it back with every event. Exporter hot path becomes:

```rust
let event = decoder.decode(record)?;
let id = *my_map.entry(event.schema_id)
    .or_insert_with(|| exporter.register(event.event_data.format()));
exporter.write(id, event.event_data.event_data());
```

one_collect/src/etw/tdh.rs
```diff
@@ -0,0 +1,953 @@
...
+    is_32bit: bool,
+    depth: usize,
+) -> Result<(), TdhDecodeError> {
+    for i in range {
+        if i >= properties.len() {
+            return Err(TdhDecodeError::Malformed("property index out of bounds"));
+        }
+
+        let prop = &properties[i];
+        let name = read_property_name(tei_buf, prop);
```
utpilla: Minor: `read_property_name` returns `""` for fields with no name offset. Two unnamed leaves under the same prefix end up with identical `qualified_name`s (`""`, `""` or `"Parent."`, `"Parent."`), and since `EventFormat::get_field` returns the first match, fields 2+ become unreachable through the field-lookup API which leads to silent data loss with no warning.

Suggest a small fallback:

```rust
let raw_name = read_property_name(tei_buf, prop);
let raw_name = if raw_name.is_empty() { format!("field{}", i) } else { raw_name };
```

one_collect/src/etw/tdh.rs
```diff
@@ -0,0 +1,953 @@
...
+/// Returns the type_name string for a TDH_INTYPE.
+///
+/// Delegates to [`intype_to_field_info`] and returns just the name.
+fn intype_to_type_name(in_type: i32) -> &'static str {
+    intype_to_field_info(in_type, false).0
+}
+
+// ── Tests ───────────────────────────────────────────────────────────
+
+#[cfg(test)]
```
utpilla, line 683: The unit tests here for the cache-hit path, variable strings, and end-to-end `try_get_field_data_closure` resolution are good.

Three coverage gaps worth filling before merge, ordered by importance:

1. Nested struct → dot-notation. The module docstring claims it as supported, and `walk_properties` recursion is the most complex code in the file, but no test exercises it. A TL schema with one struct property asserting `parent.child` field naming would lock the behavior in.

2. `NotFound` error path. A three-line test (record with `ExtendedDataCount = 0` → `decode` returns `Err(TdhDecodeError::NotFound))` pins down the public-API contract on that variant.

3. Pointer-width cache split. The two-map split keyed by `(schema_tl_bytes, is_32bit)` is an architectural decision worth a regression test. Same TL bytes under 32-bit and 64-bit headers should produce two distinct cache entries.

one_collect/src/etw/tdh.rs
```diff
@@ -0,0 +1,953 @@
...
+        .collect();
+    String::from_utf16_lossy(&u16s)
+}
+
+/// Maps a TDH in-type to `(type_name, LocationType, size)`.
+///
+/// This is the single source of truth for the in-type → field-info
+/// mapping, used by both `walk_properties` (schema construction) and
+/// `intype_to_type_name` (explicit-length fallback).
+fn intype_to_field_info(in_type: i32, is_32bit: bool) -> (&'static str, LocationType, usize) {
```
utpilla:
```suggestion
const fn intype_to_field_info(in_type: i32, is_32bit: bool) -> (&'static str, LocationType, usize) {
```

one_collect/src/etw/tdh.rs
```diff
@@ -0,0 +1,953 @@
...
+
+        // Unknown — treat as variable-length placeholder
+        _ => ("unsupported", LocationType::Static, 0),
+    }
+}
+
+/// Returns the type_name string for a TDH_INTYPE.
+///
+/// Delegates to [`intype_to_field_info`] and returns just the name.
+fn intype_to_type_name(in_type: i32) -> &'static str {
```
utpilla:
```suggestion
const fn intype_to_type_name(in_type: i32) -> &'static str {
```

one_collect/src/etw/tdh.rs
swashtek: Replaced `is_new_schema` with `schema_id`.

utpilla (approved):
Thanks @swashtek for working on this and for your patience with the review!

beaubelgrave (approved):
Thanks for taking this work on, looks good!

utpilla:
Addresses #270

PR #280, opened 2026-06-02
Tighten visibility:  →  for crate-internal items

utpilla:
## Summary

Converts 134 items from `pub` to `pub(crate)` across `one_collect` (132) and `ruwind` (2), as flagged by rustc's `unreachable_pub` lint. No behavior changes, no public API changes.

## Motivation

In Rust, `pub` means "visible in the parent module's scope," not "visible to the world." When `pub fn foo()` lives in a `mod bar;` (private module), the `pub` is misleading. `foo` is unreachable from outside the crate regardless. The honest spelling is `pub(crate)`.

The candidates were generated mechanically with:

```bash
cargo clippy --lib -- -W unreachable_pub -A clippy::all
```

Tightening these has three benefits:

1. **Accurate intent at the call site.** A reader seeing `pub fn` reasonably assumes it's part of the published API. `pub(crate) fn` says exactly what it is.
2. **Honest visibility in tooling.** API-surface analyzers (`cargo public-api`, semver checkers, `rustdoc --document-private-items`) and IDE summaries treat `pub` and `pub(crate)` distinctly. After this change they reflect the true external surface.
3. **Refactoring safety.** If `mod bar;` were ever changed to `pub mod bar;`, all the `pub` items inside would silently become external API. `pub(crate)` prevents that accidental expansion.

### DIFF
--- a/one_collect/Cargo.toml
+++ b/one_collect/Cargo.toml
@@ -67,6 +67,9 @@ harness = false
 name = "kallsyms"
 harness = false
 
+[lints.rust]
+unreachable_pub = "deny"
+
 [lints.clippy]
 absurd_extreme_comparisons = "deny"
 almost_swapped = "deny"
--- a/one_collect/examples/perf_cpu.rs
+++ b/one_collect/examples/perf_cpu.rs
@@ -36,7 +36,7 @@ impl Utilization {
     }
 
     #[cfg(target_os = "linux")]
-    pub fn new(session: &mut PerfSession) -> Writable<Self> {
+    fn new(session: &mut PerfSession) -> Writable<Self> {
         let util = Writable::new(Self::create());
 
         /* CPU is ancillary data for perf_event_open */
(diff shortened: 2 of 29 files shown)


brianrob:
This seems like goodness to me.  I'm wondering if we should add this to each crate:

```
[lints.rust]
 unreachable_pub = "deny"
```

This would enforce the tightening that you're doing here in the future.

utpilla:
> This seems like goodness to me. I'm wondering if we should add this to each crate:
> 
> ```
> [lints.rust]
>  unreachable_pub = "deny"
> ```
> 
> This would enforce the tightening that you're doing here in the future.

@brianrob I'm in favor of doing that. I have applied your suggestion in https://github.com/microsoft/one-collect/pull/280/changes/63b00f75e70c92bcfeb9ce7b147a9b529ad5d11c. This led to identification of a few more unreachable pubs.

brianrob (approved):
LGTM.  Thanks!

PR #284, opened 2026-06-08
Add ETW/TDH decoder integration tests and gate them on PR CI

utpilla:
Related to #270 

## Summary

Adds integration tests for TDH decoding and wires it into PR CI as a required check. The tests emit real ETW TraceLogging events end-to-end and assert that `EtwSession` + `TdhDecoder` produce the expected schema and field values, giving us regression coverage for the TDH decoder shipped in #278.

### DIFF
--- a/one_collect/tests/etw_tdh_integration.rs
+++ b/one_collect/tests/etw_tdh_integration.rs
@@ -1,0 +1,938 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+//! End-to-end integration tests for the runtime TDH decoder.
+//!
+//! These tests register a real TraceLogging ETW provider (using the
+//! `tracelogging` and `tracelogging_dynamic` crates), emit known events
+//! through ETW, capture them inside an [`EtwSession`], decode them with
+//! [`TdhDecoder`], and assert that the decoded field values match what
+//! was written.
+//!
+//! Both tests are marked `#[ignore]` because the consumer side of ETW
+//! requires administrative privileges (`SeSystemProfilePrivilege` /
+//! `SeDebugPrivilege`).  Run manually from an elevated shell with:
+//!
+//! ```text
+//! cargo test -p one_collect --test etw_tdh_integration -- --ignored --nocapture --test-threads=1
+//! ```
+//!
+//! `--test-threads=1` is required: each test starts its own ETW kernel
+//! consumer session and registers the same TraceLogging provider GUID
+//! in this process.  Running the tests concurrently would race on those
+//! process- and system-global resources.
+
+#![cfg(target_os = "windows")]
+
+use std::cell::RefCell;
+use std::pin::Pin;
+use std::rc::Rc;
+use std::sync::Arc;
+use std::sync::atomic::{AtomicUsize, Ordering};
+use std::time::{Duration, Instant};
+
+use one_collect::{Guid, Writable};
+use one_collect::etw::{EtwSession, LEVEL_VERBOSE};
+use one_collect::etw::tdh::TdhDecoder;
+use one_collect::event::Event;
+use one_collect::event::os::windows::WindowsEventExtension;
+
+use tracelogging as tlg;
+use tracelogging_dynamic as tld;
+
+// ── Test fixtures ────────────────────────────────────────────────────
+
+/// Keyword used by every event the tests emit (and the value the wide
+/// event subscribes to via `MatchAnyKeyword`).
+const TEST_KEYWORD: u64 = 0x1;
+
+/// Hard upper bound on a single test run — should never be reached when
+/// running on a healthy ETW subsystem.
+const TEST_TIMEOUT: Duration = Duration::from_secs(30);
+
+/// Expected u32 value in "EventNumbers".
+///
+/// `0xDEADBEEF` is chosen deliberately: every byte is distinct and the
+/// value is asymmetric, so any byte-order or offset-by-one regression
+/// in the decoder will fail the assertion (a palindromic or
+/// repeated-byte constant could mask such bugs).  Do not change this
+/// value without keeping that property.
+const EXPECTED_U32: u32 = 0xDEADBEEF;
+/// Expected u64 value in "EventNumbers".
+///
+/// Chosen with the same byte-distinctness property as `EXPECTED_U32`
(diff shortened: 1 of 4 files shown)


one_collect/tests/etw_tdh_integration.rs
```diff
@@ -0,0 +1,921 @@
...
+/// Keyword used by every event the tests emit (and the value the wide
+/// event subscribes to via `MatchAnyKeyword`).
+const TEST_KEYWORD: u64 = 0x1;
+
+/// Hard upper bound on a single test run — should never be reached when
+/// running on a healthy ETW subsystem.
+const TEST_TIMEOUT: Duration = Duration::from_secs(30);
+
+/// Expected u32 value in "EventNumbers".
+const EXPECTED_U32: u32 = 0xDEADBEEF;
```
swashtek, line 60: Is `EXPECTED_U32` chosen deliberately here? 
I understand that we want to catch mismatch in byte order while decoding.
Also, worth noting the intent to avoid changing it.

one_collect/tests/etw_tdh_integration.rs
```diff
@@ -0,0 +1,921 @@
...
+//! through ETW, capture them inside an [`EtwSession`], decode them with
+//! [`TdhDecoder`], and assert that the decoded field values match what
+//! was written.
+//!
+//! Both tests are marked `#[ignore]` because the consumer side of ETW
+//! requires administrative privileges (`SeSystemProfilePrivilege` /
+//! `SeDebugPrivilege`).  Run manually from an elevated shell with:
+//!
+//! ```text
+//! cargo test -p one_collect --test etw_tdh_integration -- --ignored --nocapture
```
swashtek: The CI comment and the example above disagree on `--test-threads=1`. 
Is it deliberate when running manually? Otherwise, tests can run in parallel IMO.

one_collect/tests/etw_tdh_integration.rs
```diff
@@ -0,0 +1,921 @@
...
+            // EventScalars: f32 + f64 + bool8 + FILETIME + SYSTEMTIME + GUID.
+            //
+            // `add_u8(name, value, OutType::Boolean, 0)` is the dynamic
+            // equivalent of the static `bool8(name, &val)` macro keyword:
+            // both produce `InType::U8` with `OutType::Boolean`.
+            builder
+                .reset("EventScalars", tlg::Level::Verbose, TEST_KEYWORD, 0)
+                .add_f32("Pi", EXPECTED_F32, tlg::OutType::Default, 0)
+                .add_f64("E", EXPECTED_F64, tlg::OutType::Default, 0)
+                .add_u8("Flag", EXPECTED_BOOL8, tlg::OutType::Boolean, 0)
```
swashtek, line 627: For `bool8` field, TDH reports `InType = TDH_INTYPE_UINT8`, the decoder maps to `u8` and `OutType::Boolean` is silently ignored. Test schema correctly matches it. Future OutType aware change would need the assertion update. Is it intended? 

Another question on same line is that the `TDH_INTYPE_BOOLEAN` is `bool32` (4 bytes, same as TraceLogging), but it's mapped to size `1`. Is it intended that test only checks `bool8` ?
Worth adding `bool32` case to scalars event. Wdyt?

one_collect/tests/etw_tdh_integration.rs
```diff
@@ -0,0 +1,921 @@
...
+            Some(r) => r,
+            None => return Ok(()),
+        };
+
+        let mut decoder = decoder.borrow_mut();
+
+        // TraceLogging providers shouldn't emit non-TraceLogging events,
+        // but be defensive: ignore decode errors rather than failing the
+        // session callback (which would propagate and stop processing).
+        if let Ok(decoded) = decoder.decode(record) {
```
swashtek: I get the intent here that we don't abort the session on stray event. Could we log the error on the `Err` path while returning `Ok(())` so the session continues? 
Something like: 

```
match decoder.decode(record) {
    Ok(decoded) => on_decoded(&decoded),
    Err(e) => eprintln!("WARN: TDH decode failed for a record: {e:?}"),
}
Ok(())
```

This preserves the don't-abort behavior, surfaces the error for debugging. Wdyt?

one_collect/tests/etw_tdh_integration.rs
utpilla: Yes, this was deliberately chosen. I have added a comment so that we avoid changing it.

one_collect/tests/etw_tdh_integration.rs
utpilla: Updated the doc comment to add `--test-threads=1` for consistency. 

We want the tests to run sequentially since each test starts its own ETW kernel session with the same provider GUID, and concurrent registrations of the same provider in one process are not safe.

one_collect/tests/etw_tdh_integration.rs
utpilla: I have added the `eprintln!` statement.

one_collect/tests/etw_tdh_integration.rs
utpilla: Good catch. I investigated this and the decoder mapping at tdh.rs is actually wrong: https://github.com/microsoft/one-collect/blob/86c78bc6c76d8dedb0079538a04a6f2d9a5611e9/one_collect/src/etw/tdh.rs#L631

`TDH_INTYPE_BOOLEAN` is mapped to `("u8", Static, 1)` but per the TDH spec it's a Win32 `BOOL` (4 bytes). Adding a `bool32` test now would either fail or silently corrupt the offsets of fields after it.

Since this PR is intentionally test-only ("no production code touched"), I'd like to keep the scope clean and split the decoder fix + bool32 coverage into a follow-up PR.

On the related OutType-aware question (`bool8` showing as `u8`): agreed, the schema assertion would need updating when OutType awareness lands, but that's a separate decoder evolution too.

swashtek:
Looks good to me. 
Thanks for adding the integration tests.

beaubelgrave (approved):
Looks good, more testing is always a good thing!

one_collect/tests/etw_tdh_integration.rs
utpilla: @swashtek Please review #286 which fixes this issue.

PR #286, opened 2026-06-10
Fix TDH decoding for Win32 BOOL

utpilla:
Follow-up to https://github.com/microsoft/one-collect/pull/284#discussion_r3383852759

## Summary

`TDH_INTYPE_BOOLEAN` is a Win32 `BOOL`: a 4-byte signed integer, not a 1-byte boolean. The runtime TDH decoder in `tdh.rs` was sharing a match arm with `TDH_INTYPE_UINT8` and incorrectly returning `("u8", Static, 1)` for it. Because field offsets in `EventFormat` are computed by accumulating declared field sizes, an undersized `BOOLEAN` field would shift every subsequent field in the same event by 3 bytes, silently corrupting all downstream reads.

This PR splits the match arm so `TDH_INTYPE_BOOLEAN` gets its own correct `("u32", Static, 4)` mapping, and extends the integration-test suite to exercise the corrected path end-to-end.

### DIFF
--- a/one_collect/src/etw/tdh.rs
+++ b/one_collect/src/etw/tdh.rs
@@ -628,11 +628,17 @@ const fn intype_to_field_info(in_type: i32, is_32bit: bool) -> (&'static str, Lo
     match in_type {
         // Fixed-size scalars
         TDH_INTYPE_INT8                          => ("s8",   LocationType::Static, 1),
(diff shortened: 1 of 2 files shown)


beaubelgrave (approved):
LGTM

PR #288, opened 2026-06-15
Make ruwind a private module of one_collect

brianrob:
## Summary

Folds the `ruwind` unwinding crate into `one_collect` as a private internal
module, so the repository builds and ships a single library crate instead of two.

## What changed

- Move `ruwind/src/*` → `one_collect/src/ruwind/` (`lib.rs` → `mod.rs`),
  preserving git history, and fix the internal module paths.
- Keep the unwinder **internal**: only the types that appear in `one_collect`'s
  public API (`ModuleKey`, `UnwindType`, `ElfLoadHeader`) are re-exported via a
  small public facade module, `one_collect::unwind`.
- Update internal references from `ruwind::` to `crate::ruwind::`, and the
  export benchmark to `one_collect::unwind`.
- Merge `ruwind`'s dependencies (`cpp_demangle`, `rustc-demangle`) into
  `one_collect` and drop the `ruwind` path dependency.
- Move `ruwind`'s test assets / `stack_gen` helper under `one_collect`.
- Remove the `ruwind` build/test/clippy steps from CI; update `CONTRIBUTING.md`
  and the design docs to reflect the new layout.

## Public API impact

No functional API change. The unwinder types that were previously reachable as
`ruwind::*` are now reachable as `one_collect::unwind::{ModuleKey, UnwindType,
ElfLoadHeader}`; the rest of the unwinder is now an internal implementation
detail.


beaubelgrave (approved):
Looks good!

PR #291, opened 2026-06-15
Record-Trace: Enable duration and memory limits

### LINKED ISSUE #281: Add duration support in record-trace
In the `record-trace` tool, need to have the ability to declare the max duration (in seconds) of a trace collection.

### LINKED ISSUE #282: Add max memory option in record-trace
In the `record-trace` tool, need to have the ability to declare the maximum amount of memory that a trace collection can perform.

beaubelgrave:
We have scenarios where record-trace should automatically stop as if a user hit control-c when either a duration or memory limit is hit.

Add --duration and --max-memory arguments to allow these scenarios.

Honor duration within the parse_until callback directly.

Spawn a thread for memory monitoring, using OS specific functions. This ensures that memory monitoring does not affect the consumer threads ability to keep up with data rates.

Tested on both Windows and Linux manually for both --duration and --max-memory arguments, ensuring they stopped the trace at the correct points.

### DIFF
--- a/record-trace/Cargo.lock
+++ b/record-trace/Cargo.lock
@@ -324,6 +324,7 @@ dependencies = [
  "one_collect",
  "tracing",
  "tracing-subscriber",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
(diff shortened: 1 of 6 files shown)


brianrob (approved):
LGTM.  Thanks!

PR #293, opened 2026-06-16
Implement `DynRelative` / `DynAbsolute` field resolution

utpilla:
## Changes

Replaces the two `todo!()` arms in `EventFormat::get_data_with_offset_direct` with a real implementation that resolves Linux ftrace dynamic-location fields (`__rel_loc` / `__data_loc`). Previously, reading such a field would panic and abort the process; it now returns the correct payload slice, or `EMPTY` on a malformed/out-of-bounds descriptor, consistent with every other arm in the function.

### DIFF
--- a/one_collect/src/event/mod.rs
+++ b/one_collect/src/event/mod.rs
@@ -588,12 +588,45 @@ impl EventFormat {
                 &slice[2..2+bytes]
             },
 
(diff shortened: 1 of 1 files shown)


beaubelgrave (approved):
Looks good, thanks!

PR #294, opened 2026-06-16
Add  method for per-CPU ring draining

utpilla:
## Changes

Adds `parse_for_cpu(cpu)` (Option 1 from #254) - drains and dispatches one CPU's ring, firing event callbacks like `parse_all` but scoped to a single CPU. This is the final piece of the wakeup-watermark work, completing `with_wakeup_watermark()` (#276) and `perf_fds()` (#279).

## Motivation

A multi-CPU `PerfSession` registers one perf fd per CPU with the consumer's event loop (epoll / `AsyncFd`). The readiness signal is per-CPU, but `parse_all` drains every ring - so a wakeup on one CPU pulled events from CPUs whose watermark hadn't tripped. `parse_for_cpu` matches the drain to the signal: one fd fires, one ring drains.

### DIFF
--- a/one_collect/src/perf_event/rb/source.rs
+++ b/one_collect/src/perf_event/rb/source.rs
@@ -533,6 +533,7 @@ pub struct RingBufDataSource {
     hard_page_faults_builder: Option<RingBufBuilder<PageFaults>>,
     next_time: Option<u64>,
     oldest_cpu: Option<usize>,
+    cpu_to_reader_index: Vec<Option<usize>>,
     in_process_ring_buf: Option<InProcessRingBuf>,
 }
 
@@ -566,6 +567,7 @@ impl RingBufDataSource {
             hard_page_faults_builder,
             next_time: None,
             oldest_cpu: None,
+            cpu_to_reader_index: Vec::new(),
             enabled: false,
             in_process_ring_buf: None,
         }
@@ -675,6 +677,8 @@ impl RingBufDataSource {
             None => { &empty_pids },
         };
 
+        self.cpu_to_reader_index.resize(cpu_count() as usize, None);
+
         /* Build the kernel only dummy rings first */
         for i in 0..cpu_count() {
             let mut cpu_buf = common.for_cpu(i);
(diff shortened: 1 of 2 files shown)


one_collect/src/perf_event/rb/source.rs
```diff
@@ -533,6 +533,7 @@ pub struct RingBufDataSource {
     hard_page_faults_builder: Option<RingBufBuilder<PageFaults>>,
     next_time: Option<u64>,
     oldest_cpu: Option<usize>,
+    cpu_reader_index: HashMap<u32, usize>,
```
beaubelgrave: Since we map index to CPU, lets just use a Vec here that expands to the amount of index's we have. This will speed up the lookup compared to a hash.

one_collect/src/perf_event/rb/source.rs
```diff
@@ -1271,6 +1278,78 @@ impl PerfDataSource for RingBufDataSource {
         self.enabled
     }
 
+    fn begin_reading_cpu(
```
beaubelgrave, line 1283: It would be nice to have begin_reading_cpu() return a struct that contains the reader, cursor that then is passed to read_cpu(). This would eliminate several branches and memory lookups within read_cpu() that is called in a tight loop when on a per-CPU looping path.

beaubelgrave: It could also contain a copy of the ancillary as well.

one_collect/src/perf_event/rb/source.rs
utpilla: Fixed in https://github.com/microsoft/one-collect/pull/294/changes/2d7ffc3a37af6d188035071a3fadfbed7d88c860.

one_collect/src/perf_event/rb/source.rs
utpilla: > It would be nice to have begin_reading_cpu() return a struct that contains the reader, cursor that then is passed to read_cpu(). This would eliminate several branches and memory lookups within read_cpu() that is called in a tight loop when on a per-CPU looping path.

Addressed the lookup part in https://github.com/microsoft/one-collect/commit/a0bd0cde5121a139d1f02d40cd0910324cbc9aa0. `begin_reading_cpu()` now returns a `CpuReadContext` passed to `read_cpu()`/`end_reading_cpu()`, so the loop resolves the reader slot once and drops the per-record lookup and `Option` match.

I didn't optimize away the bounds checks, so the `readers[i]`/`cursors[i]` indexing is still there (same as the existing all-CPU `read()` path). We could optimize that too, but it's less straightforward and would likely involve `unsafe` code.

beaubelgrave (approved):
Thanks, looks good!

Issue #289, opened 2026-06-15
ETW: Hardcoded session GUID in `WNODE_HEADER::default()` blocks concurrent sessions and persists across crashed processes


utpilla:
## Summary

`one_collect/src/etw/abi.rs` always stamps every `EVENT_TRACE_PROPERTIES` passed to `StartTraceW` with the same hardcoded session GUID (`00000000-0000-0000-0000-000000000123`) and sets `WNODE_FLAG_TRACED_GUID`. Because ETW enforces system-wide uniqueness of the trace GUID when that flag is set, any two live ETW sessions started through this crate collide on `StartTraceW` with `ERROR_ALREADY_EXISTS (183)`  regardless of whether they are in the same process or in different processes, and regardless of the session *name* chosen by each caller. The cross-process case is the most damaging because a crashed or killed process leaves the GUID owned by a kernel-resident session that survives until reboot or an explicit `logman stop … -ets`.

## Where

https://github.com/microsoft/one-collect/blob/a242d3db8deadc216a4ca28adbd7a3f866169b58/one_collect/src/etw/abi.rs#L204-L216

This `WNODE_HEADER` is embedded in `EVENT_TRACE_PROPERTIES::default()` and flows directly into `StartTraceW` via `TraceSession::start`.

## Why ETW rejects it

When `WNODE_FLAG_TRACED_GUID` is set, `StartTraceW` requires `Wnode.Guid` to be unique among currently running sessions. If another (or a stale) session already owns the GUID, `StartTraceW` returns `ERROR_ALREADY_EXISTS (183)` even if the session *name* is different. ETW real-time sessions being kernel objects persist across process exit; they are only released by an explicit `ControlTrace(STOP)` or a reboot.

## Observed impact

The underlying problem is simple: *no two live ETW sessions started through this crate can coexist.* That plays out in three cases:

1. **Same process, sequential: works.** `TraceSession::drop` calls `ControlTraceW(STOP)`, releasing the GUID before the next `StartTraceW`. This is why the bug went unnoticed: the common case happens to work.
2. **Same process, concurrent: broken.** Two live `TraceSession` instances both request GUID `…0123`; the second `StartTraceW` returns `ERROR_ALREADY_EXISTS (183)`. The CI check for integration test works around this by serializing tests with `--test-threads=1` (`.github/workflows/ci.yml`). No consumer can run two `EtwSession`s in the same process.
[...]

PR #296, opened 2026-06-17
Add `--session-name` flag to record-trace

utpilla:
Addresses https://github.com/microsoft/one-collect/issues/289#issuecomment-4712821835

## Changes

Adds an optional `--session-name <name>` flag to `record-trace` that sets the ETW/perf session name used for recording. Defaults to `"record-trace"` when omitted, preserving current behavior.

## Motivation

`record-trace` previously hardcoded the session name `"record-trace"`. On Windows, after the GUID-from-name fix (#289), the session name determines the session's identity:

- A fixed name (the default) means one `record-trace` ETW session per machine at a time, but a new run automatically reclaims a session left behind by a crashed one.
- A unique name (e.g. per-PID) lets multiple `record-trace` ETW sessions run side by side.

On Linux the name has no such constraint. perf sessions are scoped per PID/CPU, so concurrent runs already work regardless of name. The flag is primarily a Windows concurrency control and reads as a harmless session label on Linux.

This flag lets users opt into concurrent recording on Windows when they need it, while keeping the safe singleton + crash-recovery default for the common case.

### DIFF
--- a/record-trace/engine/src/commandline.rs
+++ b/record-trace/engine/src/commandline.rs
@@ -63,6 +63,9 @@ struct Args {
 
     #[arg(long, default_value_t = LogMode::File, help = "Log mode: 'disabled', 'console', or 'file'")]
     log_mode: LogMode,
+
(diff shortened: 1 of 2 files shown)


beaubelgrave (approved):
Thanks!

PR #297, opened 2026-06-17
Add support for emitting cgroup IDs from perf events

### LINKED ISSUE #174: Need to bubble up cgroup id from events
When tracing particular trace points (with little to no filtering), we need to have the ability to disambiguate processes that have the same name by having their cgroup id being emitted as well.

Take the following example scripts...
```
let ev_fsync = event_from_tracefs("syscalls", "sys_enter_fsync");
let ev_sync = event_from_tracefs("syscalls", "sys_enter_sync");

record_event(ev_fsync);
record_event(ev_sync);
```
and output
```
+7.3156: syscalls/sys_enter_fsync(less, PID=46649): 1 Count
Record: common_type=1012 common_flags= common_preempt_count= common_pid=46649 __syscall_nr=74 fd=3 0
+10.7635: syscalls/sys_enter_fsync(less, PID=46659): 1 Count
Record: common_type=1012 common_flags= common_preempt_count= common_pid=46659 __syscall_nr=74 fd=3 0
```

Having the cgroup id would allow us to determine who spawned the less command(s)

leafcompost:
This patch makes it possible to record the cgroup ID that perf reports on each tracepoint sample. To do so, we add a builder, `with_cgroup_data()`, that sets `PERF_SAMPLE_CGROUP` on the session (if we confirm that it has kernel support via probing). The cgroup ID is then surfaced as an attribute (`CgroupAttributeSource`) and printed by record-trace.

`PERF_SAMPLE_CGROUP` [has been part of perf_event_open(2)](https://manpages.debian.org/testing/manpages-dev/perf_event_open.2.en.html#PERF_SAMPLE_CGROUP) since Linux 5.7. When it is set, the kernel records the cgroup ID of the sampled task on every sample. one-collect [already defines](https://github.com/microsoft/one-collect/blob/598ac7172004b9c13f0412d96befe3d3986ad81d/one_collect/src/perf_event/abi.rs#L67) the PERF_SAMPLE_CGROUP constant in its ABI definitions, but it never set the bit or read the value.

For our use cases, we need to attribute each event back to the service that caused it to fire. Today we can do that attribution on the PID from the event, but PIDs get reused and the PIDs of forked processes/children have to be looked up again. With this change, each event would carry the cgroup id of the originating task, so it would let us attribute the connection to the cgroup (of its systemd service)easily.

Closes #174.

## Usage

This behavior is enabled by default. However, unlike the other optional sample fields, `with_cgroup_data()` can fail outright on kernels before 5.7, so we request it only after a kernel-support probe succeeds. Before turning it on for a session, we open a throwaway perf event with just that sample bit set. If the kernel accepts it, we obtain the cgroup ID on every sample.

A consumer can read it off the sample with `context.cgroup_id()`. Or, for a usual record-trace script like so:

```rust
let ev = event_from_tracefs("syscalls", "sys_enter_write");
record_event(ev);
```

Running record-trace in the background gives you something like this:

> +1.0023: syscalls/sys_enter_write(seq, PID=2776273): **CGroup=6533** 1 Count
> +1.0024: syscalls/sys_enter_write(seq, PID=2776273): **CGroup=6533** 1 Count

## Performance implications

The change adds very little to the per-sample hot path. When cgroup capture is on, each tracepoint sample does one extra 8-byte read at a known offset in the sample record.

To measure how much CPU time that this would add to record_trace, I start with a small program to generate some syscall traffic (here I do a million `write(2)`s):

```c
#include <fcntl.h>
#include <stdlib.h>
#include <unistd.h>

int main(int argc, char **argv)
{
    long writes = (argc > 1) ? atol(argv[1]) : 1000000L;
    double delay = (argc > 2) ? atof(argv[2]) : 1.0;

    usleep((useconds_t)(delay * 1e6));

    int fd = open("/dev/null", O_WRONLY);
    char b = 'x';
    for (long i = 0; i < writes; i++) {
        if (write(fd, &b, 1) < 0) {
            return 1;
        }
    }
    return 0;
}
```
Then, I use a script like the following to repeatedly invoke trials of record-trace, against either the main branch build or my feature branch build. The time that it measures is off of record-trace's own PID:

```python
def trial(binary):
    gen = subprocess.Popen(["/tmp/perf_gen", "1000000", "1.0"])
    time.sleep(0.2)
    rt = subprocess.Popen(
        [binary, "--script-file", "/tmp/perf.script", "--pid", str(gen.pid),
        "--out", "/tmp/rtout", "--log-mode", "disabled"],
        stdout=subprocess.DEVNULL,
        stderr=subprocess.DEVNULL,
    )
    gen.wait()
    time.sleep(0.3)
    os.kill(rt.pid, signal.SIGINT)
    _, _, usage = os.wait4(rt.pid, 0)
    return usage.ru_utime + usage.ru_stime
```

After 30 interleaved trials each:

| Version | n   | mean    | sd     | median |
| ------- | --- | ------- | ------ | ------ |
| main    | 30  | 0.1439s | 0.0233 | 0.1396 |
| after   | 30  | 0.1295s | 0.0231 | 0.1248 |

We are actually marginally faster than main (Δ = -0.014s, -~10%), because a `default_attributes` bug fix eliminates a per-sample wrapper allocation that was costing cycles on main.

### DIFF
--- a/one_collect/src/helpers/exporting/process.rs
+++ b/one_collect/src/helpers/exporting/process.rs
@@ -333,6 +333,7 @@ pub struct ExportProcessSample {
     callstack_id: u32,
     record_id: u32,
     attributes_id: u32,
+    cgroup_id: u64,
 }
 
 impl ExportProcessSample {
@@ -356,6 +357,7 @@ impl ExportProcessSample {
             callstack_id,
             record_id: 0,
             attributes_id: 0,
+            cgroup_id: 0,
         }
     }
 
@@ -396,6 +398,10 @@ impl ExportProcessSample {
         attributes_id: usize) {
         self.attributes_id = attributes_id as u32;
     }
+
+    pub fn cgroup_id(&self) -> u64 { self.cgroup_id }
+
+    pub fn set_cgroup_id(&mut self, id: u64) { self.cgroup_id = id; }
 }
 
 const EXPORT_PROCESS_FLAG_CREATED: u8 = 1 << 0;
--- a/one_collect/src/helpers/exporting/process.rs
+++ b/one_collect/src/helpers/exporting/process.rs
@@ -333,7 +333,6 @@ pub struct ExportProcessSample {
     callstack_id: u32,
     record_id: u32,
     attributes_id: u32,
-    cgroup_id: u64,
 }
 
 impl ExportProcessSample {
@@ -357,7 +356,6 @@ impl ExportProcessSample {
             callstack_id,
             record_id: 0,
             attributes_id: 0,
-            cgroup_id: 0,
         }
     }
 
@@ -398,10 +396,6 @@ impl ExportProcessSample {
         attributes_id: usize) {
         self.attributes_id = attributes_id as u32;
     }
-
-    pub fn cgroup_id(&self) -> u64 { self.cgroup_id }
-
-    pub fn set_cgroup_id(&mut self, id: u64) { self.cgroup_id = id; }
 }
 
 const EXPORT_PROCESS_FLAG_CREATED: u8 = 1 << 0;
--- a/record-trace/engine/src/recorder.rs
+++ b/record-trace/engine/src/recorder.rs
@@ -222,6 +222,11 @@ impl Recorder {
                     context.comm_name(),
                     context.pid());
 
+                let cgroup_id = context.cgroup_id();
+                if cgroup_id != 0 {
+                    let _ = write!(line, "CGroup={} ", cgroup_id);
+                }
+
                 match context.sample().value() {
                     MetricValue::Count(count) => {
                         append_count(count, &mut line);
--- a/record-trace/engine/src/recorder.rs
+++ b/record-trace/engine/src/recorder.rs
@@ -94,7 +94,8 @@ impl Recorder {
         let mut settings = ExportSettings::default()
             .with_version_attributes()
             .with_trace_context_attributes()
-            .with_activity_id_attributes();
+            .with_activity_id_attributes()
+            .with_cgroup_attributes();
 
         // CPU sampling.
         if self.args.on_cpu() {
--- a/one_collect/src/helpers/exporting/attributes.rs
+++ b/one_collect/src/helpers/exporting/attributes.rs
@@ -333,6 +333,53 @@ impl ExportAttributeSource for ActivityIdAttributeSource {
     }
 }
 
+#[derive(Default)]
+pub struct CgroupAttributeSource {
+    attributes_cache: HashMap<u64, usize>,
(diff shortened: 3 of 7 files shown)


one_collect/src/helpers/exporting/process.rs
```diff
@@ -333,6 +333,7 @@ pub struct ExportProcessSample {
     callstack_id: u32,
     record_id: u32,
     attributes_id: u32,
+    cgroup_id: u64,
```
beaubelgrave: This pattern causes the sample struct to constantly increase in size. To prevent this, we moved to the attributes pattern, where groups of unique attributes are put together and cached.

See VersionOpCodeAttributeSource in exporting/attributes.rs for an example.

These get pulled in via the with_*_attributes() pattern in the ExportSettings builder. Please add a with_cgroup_attributes() pattern that utilizes a new AttributeSource for CGROUP data. This avoids an extra 8-bytes on Windows per-sample and allows us to scale to having many attributes on a sample without additional memory costs.

It does require an extra lookup per-sample, but that is already being done on the Linux case, so I suspect the cost is low (but we should validate at a high level).

record-trace/engine/src/recorder.rs
```diff
@@ -222,6 +179,11 @@ impl Recorder {
                     context.comm_name(),
                     context.pid());
 
+                let cgroup_id = context.cgroup_id();
+                if cgroup_id != 0 {
+                    let _ = write!(line, "CGroup={} ", cgroup_id);
```
beaubelgrave, line 228: This should be pulled out from attributes once it moves to this vs a direct field (It would be great to print all attributes per-sample).

beaubelgrave (changes requested):

one_collect/src/helpers/exporting/process.rs
leafcompost: thanks! i now have it exporting through attributes.

FYI i've touched one additional function in this patch:

https://github.com/microsoft/one-collect/blob/2f4d7ea89ba368c1f22380e8e7f738ce6ae43df9/one_collect/src/helpers/exporting/mod.rs#L823-L831

that `!associated_ids.len() > 1` is erroneously evaluating to true, which causes a performance regression when i add a new attribute. i removed the `!`  on the assumption that we were going for the C construct, and i think this restores the intended fast path.

leafcompost:
actually, i'll drop https://github.com/microsoft/one-collect/pull/297/changes/2c72c627d2dbca85b5df3d2b3b0b61c020803641 (fix docs link) since #305 will go in first.

beaubelgrave (approved):
LGTM, thanks!

PR #298, opened 2026-06-19
Fix `one_collect` compilation with `--no-default-features`

utpilla:
## Problem

Building `one_collect` with `--no-default-features` fails to compile on both Linux and Windows:

```shell
error[E0432]: unresolved import `crate::helpers::dotnet::scripting`
  --> src/helpers/dotnet/os/linux.rs:34:29
   |
34 | use crate::helpers::dotnet::scripting::*;
   |                             ^^^^^^^^^ could not find `scripting` in `dotnet`
```

Windows has the same root break, plus a second latent one:

```shell
error[E0432]: unresolved import `crate::helpers::dotnet::scripting`
  --> src/helpers/dotnet/os/windows.rs:25:29

error[E0425]: cannot find type `UniversalExporter` in this scope
```

The `scripting` module is gated behind the scripting feature (on by default), but `helpers/dotnet/os/linux.rs` imported from it unconditionally. With the feature disabled the module doesn't exist, so the import can't resolve.

Gating the import alone is insufficient: three items from it (`guid_from_provider`, `event_full_name`, `DotNetProviderFlags`) are used by always-compiled code in `OSDotNetEventFactory`, so they must stay reachable without the feature.

Windows additionally gated its `UniversalExporter` import behind `#[cfg(feature = "scripting")]`, even though that type is not feature-gated (only `ScriptedUniversalExporter` is) and is used by always-compiled code in `hook_to_exporter`. With the feature disabled the type became unresolvable.

## Fix
- Moved the three shared symbols (`guid_from_provider`, `event_full_name`, `DotNetProviderFlags`) out of the feature-gated scripting module into a new ungated module `helpers/dotnet/provider`, so they're available without the `scripting` feature. The rhai-only pieces (setter methods, `CustomType` impl) stay gated.
- Updated os/linux.rs and os/windows.rs to import these from `provider` instead of `scripting::*`.
- Removed the erroneous `#[cfg(feature = "scripting")]` gate on the `UniversalExporter` import in `os/windows.rs`.
- Extended the CI "Build (no default features)" step to run on Windows as well as Linux (dropped the `if: runner.os == 'Linux'` guard) so both platforms stay guarded.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -43,7 +43,11 @@ jobs:
     - name: 'one_collect: Build'
       working-directory: one_collect
       run: cargo build --locked ${{ inputs.build-flavor == 'release' && '--release' || '' }} --verbose
-    
+
+    - name: 'one_collect: Build (no default features)'
+      working-directory: one_collect
+      run: cargo build --locked --no-default-features ${{ inputs.build-flavor == 'release' && '--release' || '' }} --verbose
+
     - name: 'one_collect: Run tests'
       working-directory: one_collect
       run: cargo test --locked ${{ inputs.build-flavor == 'release' && '--release' || '' }} --verbose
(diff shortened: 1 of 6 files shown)


brianrob (approved):
LGTM.  Thanks!

Issue #303, opened 2026-06-23
.NET process tracking keys on raw PID, so a re-used PID is permanently skipped


hoyosjs:
The Linux .NET helper discovers processes from `mmap` records and tracks them in the EventPipe (`UserEventTracker`) and perf-map (`PerfMapTracker`) workers using the **raw PID** as the identity key. Once a PID has been enabled it is recorded in the worker's `pids` map/set and is never removed for the lifetime of the trace. Because the OS reuses PIDs after a process exits, a new process that re-enters
with a PID we have already seen is silently skipped and never gets .NET user_events or perf-map symbols enabled.

## Symptoms

- A .NET process is discovered and enabled. It later exits. A new .NET process starts and the OS reuses the same PID.
- The new process emits the `dotnet_ipc_created` / `doublemapper` mmap that normally triggers tracking, but the worker's guard short-circuits it.
- Result: the re-entering process is **never enabled** - no events, no JIT/perf-map symbols - even though it is a brand-new process we have never actually tracked.
- The stale `UnixStream` to the long-gone original process also lingers in the map for the rest of the trace.

## Cause

The worker tracking state is keyed on `u32` PID alone and entries are only dropped when the worker thread ends (map/set is dropped), never per-process during the trace. PID is not a stable process identity: after a process exits the kernel is free to assign the same PID to an unrelated process. There is no identity component (process start time, or a `pidfd`) to distinguish "same process we
tracked" from "different process that happens to share the PID".

Relevant code in `one_collect/src/helpers/dotnet/os/linux.rs`:

```rust
// UserEventTracker::worker_thread_proc
let mut pids: HashMap<u32, UnixStream> = HashMap::new();
...
/* Skip if already enabled */
if pids.contains_key(&pid) {
    continue;
}
```

```rust
// PerfMapTracker::worker_thread_proc
let mut pids = HashSet::new();
...
/* Skip if already enabled */
if pids.contains(&pid) {
    continue;
}
```

## Proposed fix

Give tracked processes a stable identity instead of keying on PID alone:

- Capture the process start time at discovery (`/proc/<pid>/stat`) and store it with the entry.
- Treat a discovered PID whose recorded start time differs from the current `/proc/<pid>/stat` start time as a *new* process: evict the stale entry (closing its dead socket / disabling its old perf-map) and enable the new one.
[...]

PR #304, opened 2026-06-24
Bound diagnostic IPC I/O and worker shutdown with timeouts

### LINKED ISSUE #301: Linux: machine-wide trace can hang when debug-stopped .NET process blocks diagnostic IPC
## Symptoms

When running a machine-wide trace, if any .NET process is stopped under a debugger or is in an unhealthy state and event pipe makes no progress, tracing can hang indefinitely.

Observed behavior:
- Trace appears stuck (no progress / no command response completion)
- Ctrl+C/stop path can also hang during shutdown
- dotnet-trace using the FFI sees the same behavior

## Cause

Linux .NET diagnostic IPC reads are blocking with no read timeout.

In `one_collect/src/helpers/dotnet/os/linux.rs`, these paths call `read_exact(...)` on `UnixStream` without `set_read_timeout(...)`:
- `PerfMapContext::enable_perf_map`
- `UserEventTracker::enable_events`

If a runtime does not respond, worker threads block forever. Shutdown then blocks on `worker.join()` in `disable()`.

### Stack A: EventPipe enable worker blocked
- `Recorder::run`
- `universal.parse_until`
- `UniversalExporter::os_parse_until`
- `OSDotNetEventFactory::hook_to_exporter` build hook
- `UserEventTracker::worker_thread_proc`
- `UserEventTracker::enable_events`
- `socket.read_exact(&mut result/code/session)` **(blocks)**

### Stack B: Perf-map worker blocked
- `Recorder::run`
- `universal.parse_until`
- `UniversalExporter::os_parse_until`
- `RingBufSessionBuilder::with_dotnet_help`
- `PerfMapTracker::worker_thread_proc`
- `PerfMapContext::enable_perf_map`
- `sock.read_exact(&mut result)` **(blocks)**

### Stack C: Shutdown blocked waiting for stuck worker
- session drop path triggers `drop_event`
- dotnet drop callback calls `tracker.disable()` / `perfmap.disable()`
- `disable()` calls `worker.join()` **(waits forever if worker is blocked)**

hoyosjs:
A machine-wide trace discovers every .NET process and, per process, opens a diagnostic-IPC unix socket under `/proc/<pid>/root/tmp/...` to enable EventPipe events and perf-maps. This runs on single worker threads doing *blocking* `read_exact` / `write_all` on the socket with no timeout. If any discovered runtime is unresponsive that blocking I/O never returns. The worker wedges on the first such process, every later process is starved behind it, and shutdown (`PerfSession` drop -> `disable()` -> `worker.join()`) blocks **forever**. The whole trace session locks up.

## What this PR does

This PR now has two commits in `one_collect/src/helpers/dotnet/os/linux.rs`:

- **Commit 1: time-bound diagnostic IPC and shutdown wait.**
  - Set 5s read/write timeout on diagnostic socket I/O (`DIAG_SOCKET_TIMEOUT`).
  - Bound tracker shutdown wait to 10s (`WORKER_JOIN_TIMEOUT`) so `disable()`
    cannot wait forever.
- **Commit 2: make shutdown bounded even with multiple wedged runtimes.**
  - Add a cancellation flag so workers stop accepting queued work during
    teardown.
  - Replace helper-thread join wait with worker completion channel signaling
    (cheaper than spawning a helper thread per disable).
  - Use short teardown write timeout (`DIAG_DISABLE_TIMEOUT`, 1s) for
    `disable_perf_map`.
  - In perf-map tail cleanup, move contexts out of the mutex, disable under a
    total budget (`DIAG_DISABLE_BUDGET`, 3s), then restore contexts so
    `remove_perf_maps` can still remove files.
  - The timeouts form a 2x2 of steady-state vs teardown and per-op vs
    aggregate limits, with `WORKER_JOIN_TIMEOUT` acting as the outer backstop.

## Behavior change

- A stopped/unresponsive runtime degrades instead of hanging the trace: enablement fails after <=5s and the worker can move on.
- `disable()` remains time-bounded (<=10s wait per tracker), while workers are also prevented from consuming unbounded queued work during teardown.
- Perf-map disable tail cleanup no longer holds the shared perf-map mutex across blocking socket I/O.

## Known limitations (intentionally out of scope here)

- `UnixStream::connect` is still unbounded; only post-connect I/O is timed.
- Process tracking keys on raw PID with no per-process eviction, so a re-used PID is mishandled (#303).
- The worker is still **serial and single-shot**. A stopped process still costs the worker up to ~5s per blocking op before it moves on, and a runtime that was only *temporarily* unreachable gets a single attempt and is dropped on failure. Retrying unreachable runtimes is a possible follow-up.

Fixes https://github.com/microsoft/one-collect/issues/301

### DIFF
--- a/one_collect/src/helpers/dotnet/os/linux.rs
+++ b/one_collect/src/helpers/dotnet/os/linux.rs
@@ -14,6 +14,7 @@ use std::collections::{HashSet, HashMap};
 use std::collections::hash_map::Entry::{Occupied, Vacant};
 use std::sync::mpsc::{self, Sender, Receiver};
 use std::sync::{Arc, Mutex};
+use std::sync::atomic::{AtomicBool, Ordering};
 use std::thread::{self, JoinHandle};
 
 use crate::helpers::dotnet::*;
@@ -47,12 +48,37 @@ fn is_dotnet_memfd_mapping(filename: &str) -> bool {
     filename.starts_with("/memfd:doublemapper")
 }
 
+/* These timeouts form a 2x2 of {steady-state, teardown} x {per-op, aggregate}.
+ * Steady-state ops can legitimately take a few seconds (DIAG_SOCKET_TIMEOUT),
+ * while teardown ops are tiny single writes and should fail fast
+ * (DIAG_DISABLE_TIMEOUT). DIAG_DISABLE_BUDGET caps the teardown loop, and
+ * WORKER_JOIN_TIMEOUT is the outer backstop on disable() itself.
+ *
+ * Invariant: WORKER_JOIN_TIMEOUT should be >= the worst-case time the worker
+ * needs to reach completion after cancellation, i.e. one in-flight op plus
+ * DIAG_DISABLE_BUDGET. With a 5s in-flight op + 3s budget = 8s < 10s, the join
+ * backstop does not fire on a merely-slow-but-healthy teardown. (The events
+ * worker's enable_events can do up to three reads at DIAG_SOCKET_TIMEOUT each,
+ * which can exceed this; the join then detaches+warns, which is harmless.) */
+
 /// Conservative timeout for diagnostic socket reads and writes. Prevents
-/// indefinite blocking when a .NET runtime is stopped (e.g. under a debugger).
+/// indefinite blocking when a .NET runtime is unresponsive.
 const DIAG_SOCKET_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(5);
 
 /// Timeout for waiting on worker threads during shutdown.
 const WORKER_JOIN_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(10);
+
+/// Short timeout for diagnostic socket writes on the teardown path. A
+/// disable request is near-instant against a healthy runtime, so a wedged
+/// one fails fast instead of consuming the full read/write timeout.
+const DIAG_DISABLE_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(1);
+
+/// Overall budget for best-effort disable of tracked runtimes at shutdown.
+/// Caps total teardown time so a large number of wedged runtimes can't make
+/// shutdown scale with process count. Sized to attempt roughly three teardown
+/// writes (each bounded by DIAG_DISABLE_TIMEOUT) before giving up.
+const DIAG_DISABLE_BUDGET: std::time::Duration =
+    std::time::Duration::from_secs(3 * DIAG_DISABLE_TIMEOUT.as_secs());
 
 struct PerfMapContext {
     tmp: OpenAt,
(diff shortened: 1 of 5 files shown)


one_collect/src/helpers/dotnet/os/linux.rs
```diff
@@ -538,12 +635,22 @@ impl PerfMapTracker {
...
+        /* Wait for the worker to finish, but never block shutdown
+         * indefinitely if it is wedged in an un-interruptible syscall.
+         * The worker signals completion on the done channel; on timeout we
+         * detach (drop the handle without joining) as a backstop. */
         if let Some(worker) = self.worker.take() {
-            let _ = worker.join();
+            if self.done.recv_timeout(WORKER_JOIN_TIMEOUT).is_ok() {
+                let _ = worker.join();
+            } else {
+                warn!("oc-dotnet-perfmap worker did not exit within timeout");
```
beaubelgrave: In this case, will the perf maps still get cleaned up?

beaubelgrave (approved):
LGTM, @brianrob do you want to also look?

one_collect/src/helpers/dotnet/os/linux.rs
```diff
@@ -47,6 +48,41 @@ fn is_dotnet_memfd_mapping(filename: &str) -> bool {
...
+
+/// Short timeout for diagnostic socket writes on the teardown path. A
+/// disable request is near-instant against a healthy runtime, so a wedged
+/// one fails fast instead of consuming the full read/write timeout.
+const DIAG_DISABLE_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(1);
+
+/// Default budget for the perf-map disable loop at shutdown. Overridable via
+/// DotNetHelper::with_cleanup_timeout (record-trace: --dotnet-cleanup-timeout)
+/// to trade teardown time for a cleaner final state.
+const DEFAULT_DOTNET_CLEANUP_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(10);
```
brianrob: A scenario that I'm concerned about is one in which there are many processes - lets say 100s.  In that case, is 10 seconds enough?  The concept of individual timeouts makes sense for sure.  And even having an upper bound on the entirety of the operation makes sense.  I just wonder if 10 seconds is too small.  I'm inclined to do something like 60 seconds and allow folks to reduce that if they want.  The goal is to ensure that the system doesn't block indefinitely, but I want to make sure that we do favor clean-up of as many processes as possible while handling the case that a runtime is unresponsive.

one_collect/src/helpers/dotnet/os/linux.rs
```diff
@@ -427,12 +486,22 @@ impl UserEventTracker {
...
+        /* Wait for the worker to finish, but never block shutdown
+         * indefinitely if it is wedged in an un-interruptible syscall.
+         * The worker signals completion on the done channel; on timeout we
+         * detach (drop the handle without joining) as a backstop. */
         if let Some(worker) = self.worker.take() {
-            let _ = worker.join();
+            if self.done.recv_timeout(dotnet_worker_join_timeout(std::time::Duration::ZERO)).is_ok() {
+                let _ = worker.join();
+            } else {
+                warn!("oc-dotnet-events worker did not exit within timeout");
```
brianrob: Can you please update the log messages throughout this PR that are related to timeouts and make sure that they log the length of the timeout?

brianrob (commented):
Overall looks great - a couple of comments/questions below.

brianrob (approved):
Thanks @hoyosjs!  LGTM.

Issue #299, opened 2026-06-19
Expose ETW session loss metrics


utpilla:
## Summary

`one_collect` currently provides no way to read the health counters of a running ETW real-time session. When the kernel drops events because the buffer pool is exhausted or drops whole buffers because the real-time consumer cannot keep up, those losses are invisible to consumers of the crate. As a result, a pipeline can silently lose data with no signal.

We should expose these counters so downstream pipelines can emit them as metrics and detect silent data loss. Primary focus is **EventsLost** and **RealTimeBuffersLost**; **LogBuffersLost** and **BuffersWritten** come back at no extra cost (same struct, same single query call).

## What's missing today

ETW maintains live counters on the session's `EVENT_TRACE_PROPERTIES`, updated as the session runs, but nothing in the crate reads them back. The fields:

- **EventsLost** - individual events the kernel dropped at write time because every buffer in the pool was full and it could not allocate another within `MaximumBuffers`. Producer-side loss. Cumulative since session start.
- **RealTimeBuffersLost** - whole buffers that could not be delivered to the real-time consumer (the `ProcessTrace` loop) because it was not draining fast enough and the kernel's internal real-time delivery queue overflowed. Consumer-side loss; each lost buffer is many events.
- **LogBuffersLost** - whole buffers that could not be flushed to a log file. ~0 for real-time (non-file) sessions. Returned for completeness.
- **BuffersWritten** - total buffers written; useful as a denominator for a loss rate. Not a loss metric. Returned for completeness.

## Why a query API (not a callback)

Unlike the Linux/perf side (`PERF_RECORD_LOST` injected inline into the ring buffer, surfaced through `lost_event()` callbacks), ETW does **not** deliver these counters as event records. They are live fields read back from the session via `ControlTraceW(handle, ..., EVENT_TRACE_CONTROL_QUERY)`. So the natural shape is a query, callable while the session runs.

Two properties shape the design:
- These counters are **session-wide and cumulative**. They **cannot** be attributed to a specific provider or event - the dropped events were never recorded. Per-provider attribution would require running separate sessions per provider.
- The session handle is a bare `u64` and is `Send`, so a query can run on any thread. This matters for sessions started with a never-true `until` predicate (run-forever), where there is no "end of run" to read final totals from - live polling is the only option.

## Proposed public API

A small returned struct plus one free function and two convenience methods, mirroring the existing `flush_trace` pattern.

```rust
/// Cumulative loss/health counters for a running ETW session.
[...]

PR #306, opened 2026-07-06
Add query_stats function and TraceStats struct for ETW session metrics

swashtek:
This PR adds support for querying ETW runtime session statistics from a running session. It introduces a public `TraceStats` API, exposes the session handle through `SessionCallbackContext`, and wires the implementation to ETW’s `ControlTraceW(..., QUERY)` path

https://github.com/microsoft/one-collect/issues/299

### DIFF
--- a/one_collect/src/etw/abi.rs
+++ b/one_collect/src/etw/abi.rs
@@ -883,16 +883,16 @@ pub(super) struct TraceStatsRaw {
 pub(super) fn query_stats(handle: u64) -> anyhow::Result<TraceStatsRaw> {
     let mut properties = EVENT_TRACE_PROPERTIES::for_control();
 
-    unsafe {
-        let result = ControlTraceW(
+    let result = unsafe {
+        ControlTraceW(
             handle,
             std::ptr::null(),
             &mut properties,
-            EVENT_TRACE_CONTROL_QUERY);
+            EVENT_TRACE_CONTROL_QUERY)
+    };
 
-        if result != 0 {
-            anyhow::bail!("ControlTraceW query failed with {}", result);
-        }
+    if result != 0 {
+        anyhow::bail!("ControlTraceW query failed with {}", result);
     }
 
     Ok(TraceStatsRaw {
--- a/one_collect/src/etw/abi.rs
+++ b/one_collect/src/etw/abi.rs
@@ -873,7 +873,7 @@ impl TraceEnable {
     }
 }
 
-pub(super) fn query_stats(handle: u64) -> anyhow::Result<super::TraceStats> {
+pub(super) fn query_stats(handle: u64) -> anyhow::Result<super::SessionStats> {
     let mut properties = EVENT_TRACE_PROPERTIES::for_control();
 
     let result = unsafe {
@@ -888,7 +888,7 @@ pub(super) fn query_stats(handle: u64) -> anyhow::Result<super::TraceStats> {
         anyhow::bail!("ControlTraceW query failed with {}", result);
     }
 
-    Ok(super::TraceStats {
+    Ok(super::SessionStats {
         events_lost: properties.EventsLost,
         real_time_buffers_lost: properties.RealTimeBuffersLost,
         log_buffers_lost: properties.LogBuffersLost,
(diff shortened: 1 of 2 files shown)


one_collect/src/etw/mod.rs
```diff
@@ -274,6 +364,26 @@ impl SessionCallbackContext {
...
+    pub fn handle(&self) -> u64 { 
+        self.handle
+    }
+
+    /// Query the running session's loss counters.
+    ///
+    /// Convenience wrapper over [`query_stats`] using this context's
+    /// handle. Only valid while the session is running; see
+    /// [`query_stats`] for handle-lifecycle and error semantics.
+    pub fn query_stats(&self) -> anyhow::Result<TraceStats> {
```
utpilla: Is this method needed? Every use case should be covered by `handle()` + the free function (`query_stats(ctx.handle()`), and this method can't do anything the free function can't.

one_collect/src/etw/abi.rs
```diff
@@ -872,6 +873,36 @@ impl TraceEnable {
...
+    pub real_time_buffers_lost: u32,
+    pub log_buffers_lost: u32,
+    pub buffers_written: u32,
+}
+
+pub(super) fn query_stats(handle: u64) -> anyhow::Result<TraceStatsRaw> {
+    let mut properties = EVENT_TRACE_PROPERTIES::for_control();
+
+    unsafe {
+        let result = ControlTraceW(
```
utpilla: Could we reduce the scope of unsafe to just the call so that the unsafe boundary is minimal?

```rust
let result = unsafe {
    ControlTraceW(
        handle,
        std::ptr::null(),
        &mut properties,
        EVENT_TRACE_CONTROL_QUERY)
};

if result != 0 {
    anyhow::bail!("ControlTraceW query failed with {}", result);
}
```

utpilla (approved):
Left some minor suggestions.

one_collect/src/etw/mod.rs
```diff
@@ -51,6 +51,96 @@ pub const CAPTURE_STATE: u32 = abi::EVENT_CONTROL_CODE_CAPTURE_STATE;
...
+    /// Real-time delivery buffers lost in transit to the consumer
+    /// (widened from the native `u32`).
+    pub real_time_buffers_lost: u64,
+    /// Buffers that could not be flushed to the log file
+    /// (widened from the native `u32`).
+    pub log_buffers_lost: u64,
+    /// Buffers successfully written since session start
+    /// (widened from the native `u32`).
+    pub buffers_written: u64,
+}
```
utpilla: @swashtek I missed this when creating the issue #299 but we shouldn't be doing the `u32` -> `u64` widening in `one_collect`.

It's best to let the users of this API deal with `u32` to `u64` conversion if the need it. This would also simplify the changes here. we could just have one struct for `TraceStats`.

```rust
pub struct TraceStats {
    pub events_lost: u32,
    pub real_time_buffers_lost: u32,
    pub log_buffers_lost: u32,
    pub buffers_written: u32,
}
```

swashtek: Agreed, it's good to move this conversion on consumer side, addressed in 59fa02f56ea77334e842d49bd7dfb84b0afadb2c

one_collect/src/etw/mod.rs
```diff
@@ -51,6 +51,87 @@ pub const CAPTURE_STATE: u32 = abi::EVENT_CONTROL_CODE_CAPTURE_STATE;
...
+///
+/// All counters are cumulative since session start and reset only when
+/// the session is stopped. Apply deltas between consecutive polls to
+/// track rate-of-loss metrics.
+///
+/// The native ETW counters are 32-bit and can wrap on long-running
+/// sessions, so compute deltas using wrapping subtraction rather than
+/// assuming strict monotonicity.
+#[derive(Debug, Clone, Copy, Default, Eq, PartialEq)]
+pub struct TraceStats {
```
brianrob: Should this be called `SessionStats`?

brianrob (commented):
Overall, LGTM.  One naming question/thought below.

one_collect/src/etw/mod.rs
swashtek: Addressed in 90d745a

brianrob (approved):
LGTM.  Thanks!

Issue #283, opened 2026-06-08
Async perf paths do not wake up for in-process or mock-only data sources


utpilla:
Follow up to https://github.com/microsoft/one-collect/pull/279#discussion_r3375263738

PR #279 added `PerfSession::perf_fds()` so consumers can drive parsing from kernel fd readiness (for example `epoll(2)` or `tokio::io::unix::AsyncFd`).

For sources that do not own per-CPU perf fds (`MockData` and `InProcessRingBuf`), `perf_fds()` returns an empty list. A consumer wired purely to fd readiness will therefore never wake up on such sources.

The current contract is that the consumer picks the source and knows whether the fd-driven async pattern applies. Sources without fds are expected to be driven by `PerfSession::parse_all`, `parse_for_duration`, or `parse_until`, as they are today.

Possible future directions if a real use case appears:
- Give `InProcessRingBuf` a `std::task::Waker` slot (for example `Arc<Mutex<Option<Waker>>>` shared with the writer). The reader stashes its waker on `Poll::Pending`, and `InProcessRingBufWriter::write` calls `wake()` after publishing. This is std-only, works on any async runtime, and avoids a kernel-side wakeup object.
- Alternatively, add an eventfd (or similar) so `perf_fds()` can return a real fd for in-process sources.
- Add a timeout boundary to the async drain paths so they fall back to a periodic drain when no fds are present.

`MockData` is a test fixture, so the sync drain methods (`parse_all`, etc.) are likely the right answer there regardless.

PR #307, opened 2026-07-07
Filter perf_fds() by the builder's target_cpus

utpilla:
## Summary

`PerfSession::perf_fds()` returned a perf fd for every CPU, ignoring the builder's `target_cpus`. A session created with a CPU mask (`with_target_cpu`) would hand back fds for CPUs the caller explicitly excluded. This filters the exposed fds down to the configured CPU set.

## Motivation

Leader rings are created for every CPU because all-CPU events (kernel records, and `no_cpu_mask` events added via `add_event`) redirect into them. `perf_fds()` walked those leaders directly, so masked-out CPUs leaked into the returned list. A push-model consumer would then register fds for CPUs it never asked to watch.

### DIFF
--- a/one_collect/src/perf_event/mod.rs
+++ b/one_collect/src/perf_event/mod.rs
@@ -625,6 +625,13 @@ impl PerfSession {
     /// Borrow the per-CPU perf file descriptors backing this session,
     /// paired with their CPU index.
     ///
+    /// Only the CPUs selected by the builder's
+    /// [`RingBufSessionBuilder::with_target_cpu`] are returned (all CPUs
+    /// when no target is set). All-CPU events such as kernel records and
+    /// `no_cpu_mask` events still land on masked-out CPUs' rings and are
+    /// processed through the normal [`Self::parse_all`] path; they are
+    /// just not exposed as separately wakeable fds here.
+    ///
     /// Intended for consumers that want to wait on the perf fds directly
     /// (for example via `epoll(2)` or `tokio::io::unix::AsyncFd`) instead
     /// of relying on the built-in poll/sleep read loop. Pair this with
--- a/one_collect/src/perf_event/rb/source.rs
+++ b/one_collect/src/perf_event/rb/source.rs
@@ -896,6 +896,22 @@ impl RingBufDataSource {
         Ok(())
     }
 
+    /// CPUs whose leader-ring fd should be exposed by
+    /// [`PerfDataSource::perf_fds`]. CPUs masked out by the builder's
+    /// `target_cpus` are excluded so the exposed fds match the
+    /// configured CPU set. Yields CPUs in `leader_ids` order;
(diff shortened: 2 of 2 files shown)


brianrob:
@utpilla, can you share a bit more context on this change?  A concern here is that even when we target a set of CPUs, there are some events (e.g. mmap events) that may come from other CPUs not in the target set that we still need.  If we don't include them in `perf_fds()` then calls to `parse_all` or `parse_until` may be starved of these events unless/until one of the ring buffers in the target set hit the ready watermark.

beaubelgrave:
Some of what you are asking about @brianrob is covered in issue #283

utpilla:
@brianrob Quick context on why we filter: before this PR, `perf_fds()` returned an fd for every online CPU even when the session set `target_cpus`. So, a consumer that masks down to a subset (like a per-core session) would get an fd per CPU and register epoll/`AsyncFd` interest on CPUs it excluded. This PR makes `perf_fds()` match the configured set.

Your concern is valid though. It points at an inconsistency the filter exposed: sample events honor `target_cpus`, but kernel metadata (mmap/comm/fork) is captured on all CPUs (`build()` redirects kernel events with &None). So, metadata is all-CPU while the fds are target-only, and a purely fd-driven loop can miss cross-CPU mmap records until a target CPU trips its watermark. `parse_all` and `parse_until` are not affected since they drain every ring.

So, the real question becomes "should a `target_cpus` session receive kernel metadata from non-target CPUs?".
- If no (the session is fully scoped to its CPUs), the filter is correct and this is just a docs/expectations clarification.
- If yes, then the filter isn't the real fix. We'd need to line up the two behaviors: either collect metadata only on the target CPUs too or keep collecting it everywhere but give those rings a way to wake the fd loop.

What semantic would you lean towards?

beaubelgrave:
For me, mmap/comm/fork are handled at the Exporter level, and the Exporter does not support per-CPU fds, etc. (Someone could still wire them up if they choose to).

At the PerfSession level, that does not require mmap/comm/fork (that is an Export or Machine reconstruction state thing), it is not a problem.

I'm personally fine with PerfSession and perf_fds being per-CPU only and treating the mmap/comm/fork problem as a separate concern that only really materializes if we move Export or other Export-like scenarios to per-CPU that require them (and at that point that scenario owns the solution they want to implement for that).

beaubelgrave:
@brianrob you agree about per-CPU filtering scenarios owning mmap and other event problems? We don't use per-CPU in exporting, other places that choose to do per-CPU and need them should own that solution, IMHO.

brianrob:
Thanks for the explanation @utpilla.  Agreed with @beaubelgrave, I think it is fine to defer until/unless that scenario exists at which time the scenario will need to define a solution.

brianrob (approved):

brianrob:
@beaubelgrave just want to confirm - if you're good with this change, go ahead and hit merge (or tell me to.  :))

PR #308, opened 2026-07-08
[perf] Remove redundant per-event lookup in TDH decoding

utpilla:
## Summary
The `TdhDecoder` schema cache hit path did two `HashMap` lookups per event (hashing the 50-500 byte schema key twice). This indexes a backing `Vec` by a `Copy` handle so the lookup happens once, roughly halving the per-event cache-lookup cost. No behavior change.

## Problem
`decode()` looked the schema up with a `get().is_none()` probe followed by a `get().expect()`:

```rust
if self.cache.get(key, is_32bit).is_none() {
    // ... build + insert on miss ...
}
let schema = self.cache.get(key, is_32bit).expect("just inserted");
```

On a cache **hit** (the common case) this hashes the large key **twice**. It is a workaround for a borrow-checker limitation ([NLL Problem Case #3](https://rust-lang.github.io/rfcs/2094-nll.html#problem-case-3-conditional-control-flow-across-functions)): returning a reference borrowed from the map on one branch prevents the `&mut` insert on the other, so you cannot do get-or-insert in a single lookup and return the reference.

## Fix
Store the decoded schemas in an append-only `Vec<CachedSchema>` (`schemas`) and make the maps hold a `usize` index into it. On a hit, the index is copied out (ending the map borrow) and the schema is read from `schemas[idx]` -- a single hash, no `unsafe`, no new deps. This is the same index-by-handle pattern the Linux nettrace path already uses.

```rust
let idx = match self.cache.index_of(key, is_32bit) {
    Some(idx) => idx,                                          // hit: single hash
    None => self.cache.insert(key.to_vec(), is_32bit, schema), // miss
};
let schema = self.cache.get(idx);
```

## Performance
Median per-lookup (64 schemas, 200-byte keys, criterion):

| | time |
|---|---|
| `double_get_inline` (before) | 116.6 ns |
| `single_get_indexed` (after) | 57.7 ns |

**~2.0x faster (~51% reduction)** on the lookup, matching the "one hash vs two" expectation.

<details>
<summary>Benchmark used to measure this (criterion)</summary>

```rust
use std::collections::HashMap;
use std::hash::BuildHasherDefault;

use criterion::{black_box, criterion_group, criterion_main, Criterion};
use twox_hash::XxHash64;

type XxMap<V> = HashMap<Vec<u8>, V, BuildHasherDefault<XxHash64>>;

const SCHEMA_COUNT: usize = 64; // distinct schemas (call sites)
const KEY_LEN: usize = 200;     // representative TraceLogging schema-TL key

// Distinct keys; only a few bytes differ so hashing still reads the whole
// buffer, as it would for real schema bytes.
fn build_keys() -> Vec<Vec<u8>> {
    (0..SCHEMA_COUNT)
        .map(|i| {
            let mut k = vec![0u8; KEY_LEN];
            k[0] = i as u8;
            *k.last_mut().unwrap() = i as u8;
            k
        })
        .collect()
}

fn bench_cache_lookup(c: &mut Criterion) {
    let keys = build_keys();

    // Before: schema stored inline as the map value; the hot path hashes twice.
    let inline: XxMap<u64> = keys.iter().map(|k| (k.clone(), 0u64)).collect();

    // After: map -> index into a backing Vec; single hash + Copy index read.
    let mut index: XxMap<usize> = XxMap::default();
    let mut schemas: Vec<u64> = Vec::new();
    for k in &keys {
        index.insert(k.clone(), schemas.len());
        schemas.push(0);
    }

    // Mid-table key to avoid best/worst-case bucket bias.
    let hot = keys[SCHEMA_COUNT / 2].as_slice();
    let mut g = c.benchmark_group("tdh_cache_lookup");

    g.bench_function("double_get_inline", |b| {
        b.iter(|| {
            if inline.get(hot).is_none() {
                unreachable!("key is present");
            }
            black_box(inline.get(hot).expect("just checked")); // second hash
        });
    });

    g.bench_function("single_get_indexed", |b| {
        b.iter(|| {
            let idx = *index.get(hot).expect("key is present"); // single hash
            black_box(schemas[idx]);
        });
    });

    g.finish();
}

criterion_group!(benches, bench_cache_lookup);
criterion_main!(benches);
```

The map value's payload is trivial (`u64`) on purpose: a lookup returns a reference, so the value's size does not affect probe cost. The only difference under test is one hash vs two.

</details>

### DIFF
--- a/one_collect/src/etw/tdh.rs
+++ b/one_collect/src/etw/tdh.rs
@@ -143,28 +143,67 @@ struct CachedSchema {
 /// Hash builder using XxHash64, matching the rest of the ETW module.
 type XxBuildHasher = BuildHasherDefault<XxHash64>;
 
-/// Schema cache: two maps (32-bit / 64-bit) keyed by raw TL bytes.
+/// Schema cache: an append-only `Vec` of decoded schemas plus two maps
+/// (32-bit / 64-bit) that key raw TL bytes to an index into that `Vec`.
+///
+/// Storing a `Copy` index (rather than the schema inline in the map) lets the
+/// hot path hash the 50-500 byte key exactly once: the index is copied out,
+/// ending the map borrow, and the schema is then read from `schemas`.  The
+/// previous `get().is_none()` + `get().expect()` pattern hashed the key twice
+/// on every cache hit (a borrow-checker workaround for get-or-insert).
 struct SchemaCache {
-    cache_64: HashMap<Vec<u8>, CachedSchema, XxBuildHasher>,
-    cache_32: HashMap<Vec<u8>, CachedSchema, XxBuildHasher>,
+    /// Decoded schemas, indexed by the maps below.  Append-only: entries are
+    /// never removed, so an index stays valid for the life of the cache.
+    schemas: Vec<CachedSchema>,
+    cache_64: HashMap<Vec<u8>, usize, XxBuildHasher>,
+    cache_32: HashMap<Vec<u8>, usize, XxBuildHasher>,
 }
 
 impl SchemaCache {
     fn new() -> Self {
         Self {
+            schemas: Vec::new(),
             cache_64: HashMap::with_hasher(XxBuildHasher::default()),
             cache_32: HashMap::with_hasher(XxBuildHasher::default()),
         }
     }
 
-    fn get(&self, key: &[u8], is_32bit: bool) -> Option<&CachedSchema> {
+    /// Returns the index of a cached schema, or `None` on a miss.
+    /// The key is hashed exactly once.
+    fn index_of(&self, key: &[u8], is_32bit: bool) -> Option<usize> {
         let map = if is_32bit { &self.cache_32 } else { &self.cache_64 };
-        map.get(key)
+        map.get(key).copied()
     }
 
-    fn insert(&mut self, key: Vec<u8>, is_32bit: bool, schema: CachedSchema) {
+    /// Maps `key` to a schema index and returns it.  Intended for the cache
+    /// miss path, where `key` is not yet present.
+    ///
+    /// The schema is only pushed onto `schemas` when the key is actually
+    /// vacant, so a (contract-violating) duplicate key returns the existing
+    /// index without orphaning a freshly built schema.  The key is hashed
+    /// once via a single `entry` lookup.
+    fn insert(&mut self, key: Vec<u8>, is_32bit: bool, schema: CachedSchema) -> usize {
+        use std::collections::hash_map::Entry;
+
+        let idx = self.schemas.len();
         let map = if is_32bit { &mut self.cache_32 } else { &mut self.cache_64 };
-        map.entry(key).or_insert(schema);
+        match map.entry(key) {
+            // Already present: return the existing index and drop `schema`.
+            Entry::Occupied(e) => *e.get(),
+            Entry::Vacant(e) => {
+                let assigned = *e.insert(idx); // ends the map borrow
+                self.schemas.push(schema); // only push once actually indexed
+                assigned
+            }
+        }
+    }
+
+    /// Reads a cached schema by its index.
+    ///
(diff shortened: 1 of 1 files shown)


one_collect/src/etw/tdh.rs
```diff
@@ -143,28 +143,50 @@ struct CachedSchema {
...
-        map.get(key)
+        map.get(key).copied()
     }
 
-    fn insert(&mut self, key: Vec<u8>, is_32bit: bool, schema: CachedSchema) {
+    /// Pushes a schema, maps `key` to its index, and returns the index.
+    /// Only called on a cache miss, so `key` is not already present.
+    fn insert(&mut self, key: Vec<u8>, is_32bit: bool, schema: CachedSchema) -> usize {
+        let idx = self.schemas.len();
+        self.schemas.push(schema);
```
swashtek: `insert` function pushes `schemas` before consulting the map, then uses `or_insert(idx)`. 
What happens if future caller passes an already-present key? The pushed schema at `idx` is orphaned which siltenly discard newly built schema. THis is not the issue currently as `decode` only calls `insert` after checking `index_of` is `None`.
Wdyt of the robustness of the `insert` function? Only push when vacant or add debug log for contract violation? 
```suggestion
    fn insert(&mut self, key: Vec<u8>, is_32bit: bool, schema: CachedSchema) -> usize {
        use std::collections::hash_map::Entry;
        let idx = self.schemas.len();
        let map = if is_32bit { &mut self.cache_32 } else { &mut self.cache_64 };
        match map.entry(key) {
            Entry::Occupied(e) => *e.get(),        // already present: don't push
            Entry::Vacant(e) => {
                let assigned = *e.insert(idx);
                self.schemas.push(schema);         // only push when actually indexed
                assigned
            }
        }
    }
```

one_collect/src/etw/tdh.rs
```diff
@@ -143,28 +143,50 @@ struct CachedSchema {
...
         let map = if is_32bit { &mut self.cache_32 } else { &mut self.cache_64 };
-        map.entry(key).or_insert(schema);
+        *map.entry(key).or_insert(idx)
+    }
+
+    /// Reads a cached schema by its index.
```
swashtek, line 201:
```suggestion
    /// Reads a cached schema by its index.
    ///
    /// `idx` is always valid: it comes from `index_of`/`insert` and `schemas`
    /// is append-only (entries are never removed), so indices never dangle.
```

one_collect/src/etw/tdh.rs
swashtek, line 1154: The existing cache tests `tdh_decode_schema_cache_reuse` and `tdh_decode_pointer_width_cache_split` only have one schema per map, so a bug like "always return `schemas.last()`" would still pass silently. Please add a test that decodes distinct schemas that are interleaved and reads back to cover index schema mapping this PR introduces.

swashtek (changes requested):
Thanks for creating this change, this is well scoped perf improvement. 
I would add robustness around `insert()` and tests for distinct schemas as a safety net.

one_collect/src/etw/tdh.rs
utpilla: Addressed in https://github.com/microsoft/one-collect/pull/308/commits/0a4d6ae8a7fee4fc391a5ada964fd61269ea0bc6

one_collect/src/etw/tdh.rs
utpilla: Addressed in https://github.com/microsoft/one-collect/pull/308/commits/0a4d6ae8a7fee4fc391a5ada964fd61269ea0bc6

one_collect/src/etw/tdh.rs
utpilla: Addressed in https://github.com/microsoft/one-collect/pull/308/commits/0a4d6ae8a7fee4fc391a5ada964fd61269ea0bc6

brianrob (approved):
LGTM.  Thanks!

PR #309, opened 2026-07-12
[perf] Add `EventFormat::fields_with_data` for single-pass field decoding

utpilla:
## Summary

Adds `EventFormat::fields_with_data(data)`, an iterator that yields every field and its bytes for one event payload in a **single forward pass**. This is the efficient way to read many (or all) fields of an event with variable-length fields (strings / counted arrays).

Reading every field via the existing per-field reader (`try_get_field_data_closure`, one reader per field) is **O(n^2)** on schemas with variable-length fields: each field independently re-scans the variable-length prefix before it. `fields_with_data` walks the payload once, carrying a running offset, so each variable-length field is scanned exactly once, making a full decode **O(n)**.

## Motivation

When a schema has variable-length fields, a field's start position shifts from event to event and must be recovered by scanning the preceding variable-length fields (e.g. scanning a null-terminated string to its terminator). Resolving each field on its own re-derives that prefix from scratch every time, so decoding all `n` fields re-scans the same leading bytes once per later field: `0 + 1 + ... + (n-1) = O(n^2)`.

This cost is invisible to the existing all-fixed schemas because fixed fields sit at statically known offsets with nothing to re-scan. It only appears with variable-length fields, which is exactly the case this PR targets.

## What changed

- **New public API**: `EventFormat::fields_with_data<'a>(&'a self, data: &'a [u8]) -> impl Iterator<Item = (&'a EventField, &'a [u8])>`.
  - Fixed prefix is read at each field's absolute offset (honoring reserved gaps); once a variable-length field is reached, later fields are read relative to a running offset.
  - Counted arrays jump via their 2-byte length prefix (O(1)); null-terminated strings are scanned once.
  - A field whose length cannot be determined (and every field after it) yields an empty slice, matching what the per-field closure returns.
  - Backed by a private `FieldsWithData` iterator, so this adds **no new public types**.
- **Parity test** `event::tests::fields_with_data_matches_closures`: asserts the iterator yields byte-identical output to building and calling `try_get_field_data_closure` for each field, across all-fixed, reserved-gap, leading-strings, mid-variable, counted-array, and unreadable-size-0 layouts.
- **Integration test** (`etw_tdh_integration`): the `Message` / `Name` reads now go through `fields_with_data`, exercising the new method against real TDH-decoded ETW payloads.

`try_get_field_data_closure` stays public and unchanged for now; it is still the right tool for reading a single named field. Privatizing it is left to a follow-up PR once external consumers have migrated to `fields_with_data`.

## Benchmark

<details>
<summary>Benchmark results (per-event decode, string-prefix schema)</summary>

Each schema is `n` leading null-terminated string fields followed by a fixed `u64`.

| fields (n) | independent_per_field (O(n^2)) | single_pass (O(n)) | speedup |
|---|---|---|---|
| 4   | 271 ns  | 90 ns   | 3x   |
| 16  | 2.65 us | 291 ns  | 9x   |
| 64  | 37.3 us | 1.10 us | 34x  |
| 256 | 547 us  | 4.35 us | 126x |

As `n` quadruples: `independent_per_field` grows ~16x per step (quadratic), `single_pass` grows ~4x per step (linear). The speedup itself scales with `n` (the ratio is ~`n/2`), so wider schemas benefit more.

</details>

<details>
<summary>Benchmark code (simplified)</summary>

```rust
/// Resolve each field independently: `nth(i)` restarts the walk from the
/// front, so each field re-scans the variable-length prefix before it. O(n^2).
fn decode_independent(format: &EventFormat, data: &[u8]) {
    for i in 0..format.fields().len() {
        if let Some((field, bytes)) = format.fields_with_data(data).nth(i) {
            black_box((field, bytes));
        }
    }
}

/// Decode every field in a single forward pass. O(n).
fn decode_single_pass(format: &EventFormat, data: &[u8]) {
    for pair in format.fields_with_data(data) {
        black_box(pair);
    }
}
```

Both arms call the same `fields_with_data` internals, so the only thing measured is re-derivation (re-walking the prefix per field) vs carrying the offset forward, isolating the algorithmic difference. `nth(i)` faithfully models the old per-field reader shape (each reader independently re-walked its variable prefix).

</details>

## Follow-up
- Once no external callers remain, a follow-up PR can make `try_get_field_data_closure` crate-private.

### DIFF
--- a/one_collect/src/event/mod.rs
+++ b/one_collect/src/event/mod.rs
@@ -1314,6 +1314,45 @@ impl EventFormat {
         None
     }
 
+    /// Returns an iterator over every field and its bytes for a single event
+    /// payload, resolving all field positions in one forward pass.
+    ///
+    /// This is the efficient way to read many (or all) fields of an event: it
+    /// walks the payload once, carrying a running offset, so each
+    /// variable-length field (string / counted array) is scanned exactly once.
+    /// Reading every field is therefore O(n) in the number of fields plus the
+    /// total bytes of the variable-length fields, versus the O(n^2) cost of
+    /// resolving each field independently (where every field re-scans the
+    /// variable-length prefix before it).
+    ///
+    /// The slice yielded for each field is identical to what
+    /// [`try_get_field_data_closure`](Self::try_get_field_data_closure) would
+    /// return for that field: fields in the fixed prefix are read at their
+    /// absolute offset (honoring reserved gaps), and once a variable-length
+    /// field is reached later fields are read relative to the running offset.
+    /// A field whose length cannot be determined (and every field after it)
+    /// yields an empty slice.
+    ///
+    /// The iterator is lazy, so callers needing only the first few fields can
+    /// stop early.
+    ///
+    /// # Arguments
+    ///
+    /// * `data` - The event payload to read the field data from.
+    pub fn fields_with_data<'a>(
+        &'a self,
+        data: &'a [u8]) -> impl Iterator<Item = (&'a EventField, &'a [u8])> + 'a {
+        FieldsWithData {
+            fields: &self.fields,
+            data,
+            index: 0,
+            running_offset: 0,
+            skip_offset: 0,
+            seen_variable: false,
+            broken: false,
+        }
+    }
+
     /// Retrieves the data associated with a given `EventFieldRef` within the provided data slice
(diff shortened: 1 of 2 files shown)


brianrob (approved):
LGTM.  Thanks!

PR #311, opened 2026-07-15
Resolve Go symbols from .gopclntab for symbol-stripped Go binaries

brianrob:
## Problem
Go binaries built with `-ldflags "-s"` (e.g. kubelet, containerd, and most distro/CI Go binaries) drop the ELF symbol table, so one-collect could not resolve any function names for them — traces showed raw addresses. The Go function/line metadata is still present in the `.gopclntab` section, which the ELF symbol path did not consume.

## What this does
Adds a `.gopclntab` parser and wires it into the existing ELF symbol-resolution path so Go symbols are recovered automatically, with **no API change and no caller opt-in**. If a mapping's ELF symbols are missing/stripped, its `.gopclntab` is parsed and its functions are added as symbols.

- **`ruwind/go_pclntab.rs`** (new): parses the Go 1.18+ pclntab layouts (magic `0xfffffff0`/`0xfffffff1`), yielding `(start, end, name)` per function. Returns `None` for older/unknown magics so callers fall back cleanly.
- **`ruwind/elf.rs`**: locates `.gopclntab` (incl. PIE `.data.rel.ro`), resolves the correct text base (`runtime.text` symbol → pcHeader `textStart` → PIE relative-reloc addend → `.text`), and recovers the table by magic-scan when section headers are stripped.
- **`helpers/exporting/symbols.rs`**: `GoPclnTabSymbolReader` exposes the parsed functions as symbols.
- **`helpers/exporting/os/linux.rs` + `mappings.rs`**: integrates Go symbols into the symbol-file search; `add_matching_symbols` now transparently de-duplicates when unioning a second source onto a mapping (single-source callers pay nothing).

## Design notes
- **Cheap for non-Go binaries.** The Go path is only attempted when ELF symbols are absent, and it bails out immediately if no `.gopclntab` is found — it does not add a separate reader pass over every binary.
- **No demangling needed.** Go pclntab names are already final, human-readable form (`pkg.Type.Method`); `demangle()` returns `None`.
- **Scope:** Linux/ELF, Go 1.18+. Pre-1.18 binaries are declined by design and fall back to ELF symbols (they retain a symtab in practice).
- **Minimal public surface.** New items are `pub(crate)`/private; consumption is automatic through the existing ELF symbol path.

## Testing
- Unit tests for the parser (both magics, endianness, PIE, recovery) and two committed stripped-binary fixtures under `test/assets/go/` (self-built minimal Go programs — one symbol-stripped, one section-header-stripped). `cargo test --lib` → 128 pass; clippy clean.
- **Corpus validation** against Go's `debug/gosym` (reference implementation):
  - Synthetic matrix: **899/899 at 100.0000%** — Go 1.18/1.20/1.22/1.23 × amd64/arm64 × default/strip/PIE/cgo + section-stripped recovery.
  - Real-world: **577/580 at 100.0000%** across ~242 OSS projects (Kubernetes components, cosign, victoria-metrics, opa, terraform/vault, minio, prometheus, grafana, etc.); 3 skipped as Go <1.18 (out of scope).
  - **~41M symbols total, 0 parser defects.** `go tool objdump` independently confirmed one-collect's addresses on externally-linked/cgo binaries.


one_collect/src/ruwind/elf.rs
```diff
@@ -715,6 +717,414 @@ pub fn read_debug_link<'a>(
...
+        if value != 0 {
+            return Some(value);
+        }
+    }
+
+    let ptr_size = if pclntab.len() >= 8 { pclntab[7] as usize } else { 0 };
+
+    // 2. The pcHeader textStart field (offset 8 + 2*ptrSize). Non-zero for
+    //    internally linked binaries; zero/relocated for externally linked ones.
+    if ptr_size == 4 || ptr_size == 8 {
```
beaubelgrave, line 826: When would this be false? :)

beaubelgrave: Ah, I see. It's read from the header and we need to know if it's linked or not (and if it's a weird format or corrupt).

beaubelgrave (approved):
LGTM, that's a lot of work, thanks!

Issue #310, opened 2026-07-14
TDH Decoding Support for Manifest-Based ETW Events


utpilla:
## Summary

Extend the existing `etw::tdh::TdhDecoder` (added for TraceLogging / TraceLoggingDynamic in #270) to also decode **manifest-based** ETW events. Manifest support was explicitly deferred as future work in #270; the design there was built to accommodate it "without rework," and this issue delivers that extension.

The change reuses the entire existing decode pipeline: `TdhGetEventInformation`, the `EVENT_PROPERTY_INFO` -> `EventFormat` walker, the `LocationType` skip-chain, and the `EventData` construction. The only manifest-specific additions are (a) how a schema is *identified* for caching and (b) where the event *name* is read from.

## Motivation

The `TdhDecoder` currently handles only self-describing events, those carrying an inline `EVENT_HEADER_EXT_TYPE_EVENT_SCHEMA_TL` blob. Manifest-based providers (the bulk of Windows' own instrumentation: `Microsoft-Windows-Kernel-*`, `.NET`, countless first- and third-party providers authored with `mc.exe`) do not embed their schema in the event. Instead, the event carries only an `EVENT_DESCRIPTOR`, and the layout lives in an XML manifest registered with the OS.

`TdhGetEventInformation` is the *same* entry point for both sources, so consuming manifest events through the existing code path is a small, localized extension rather than a new subsystem. Adding it lets `one_collect` decode the large population of manifest providers through one uniform `EventFormat` / `EventData` interface.

## Background: how manifest events differ

| | TraceLogging / TraceLoggingDynamic | Manifest-based |
| --- | --- | --- |
| Where the schema lives | Inline in the event (`EVENT_HEADER_EXT_TYPE_EVENT_SCHEMA_TL`) | In an XML manifest compiled into a DLL and registered with the OS |
| Schema identity | Raw schema-TL bytes | `(Provider GUID, Id, Version)` from `EVENT_DESCRIPTOR` |
| Who registers it, where | Nobody; it's in-band | Installer/admin via `wevtutil im` (or MOF/WBEM); lands in `HKLM\...\WINEVT\Publishers` pointing at a resource DLL |
| Needed on the decoding machine | No (self-describing) | Yes, else `TdhGetEventInformation` returns `ERROR_NOT_FOUND` |
| TDH entry point | `TdhGetEventInformation` | `TdhGetEventInformation` (identical) |

[...]

PR #313, opened 2026-07-21
Enhance TDH decoder to support manifest-based events and improve schema caching logic

swashtek:
Enhance TDH decoder to support manifest-based events and improve schema caching logic

Issue: https://github.com/microsoft/one-collect/issues/310

### DIFF
--- a/one_collect/src/etw/tdh.rs
+++ b/one_collect/src/etw/tdh.rs
@@ -4,20 +4,28 @@
 //! # TDH-Based Dynamic Event Decoder
 //!
 //! This module provides [`TdhDecoder`], a runtime schema decoder for
-//! TraceLogging and TraceLoggingDynamic ETW events.  It uses the Windows
-//! Trace Data Helper (TDH) APIs to discover event schemas on the fly and
-//! converts them into the standard [`EventFormat`] / [`EventData`]
-//! representation used throughout one_collect.
+//! TraceLogging, TraceLoggingDynamic, and manifest-based ETW events.  It
+//! uses the Windows Trace Data Helper (TDH) APIs to discover event schemas
+//! on the fly and converts them into the standard [`EventFormat`] /
+//! [`EventData`] representation used throughout one_collect.
 //!
 //! ## Design
 //!
-//! The decoder caches the [`EventFormat`] directly, keyed by the raw
-//! TraceLogging schema bytes and pointer width.  Because the format uses
-//! the framework's standard `LocationType` conventions (`StaticString`,
-//! `StaticUTF16String`, etc. with `size = 0` for variable-length fields),
-//! the cached `EventFormat` is schema-stable: it doesn't depend on any
-//! particular event's payload bytes.  A cache hit collapses to a hashmap
-//! probe + `EventData::new` — effectively zero per-event overhead.
+//! The decoder caches the [`EventFormat`] directly.  TraceLogging schemas
+//! are keyed by their raw inline schema bytes; manifest schemas by their
+//! `(Provider GUID, Id, Version)` identity.  Both keyspaces are split by
+//! pointer width and index into one shared, append-only arena of decoded
+//! schemas.  Because the format uses the framework's standard `LocationType`
+//! conventions (`StaticString`, `StaticUTF16String`, etc. with `size = 0`
+//! for variable-length fields), the cached `EventFormat` is schema-stable:
+//! it doesn't depend on any particular event's payload bytes.  A cache hit
+//! collapses to a hashmap probe + `EventData::new` — effectively zero
+//! per-event overhead.
+//!
+//! Source selection is transparent: [`TdhDecoder::decode`] takes the
+//! TraceLogging path when an inline SCHEMA_TL item is present and the
+//! manifest path otherwise.  `TdhGetEventInformation` is the same entry
+//! point for both, so the byte-level decoding is entirely source-agnostic.
 //!
 //! Variable-length field resolution (scanning for null terminators, reading
 //! length prefixes) is handled lazily by the framework's existing
@@ -26,15 +34,42 @@
 //!
 //! ## Scope
 //!
-//! - **Supported**: TraceLogging and TraceLoggingDynamic events, nested
-//!   struct fields (flattened with dot-notation names), basic scalar and
-//!   string property types, 32-bit and 64-bit event payloads.
+//! - **Supported**: TraceLogging, TraceLoggingDynamic, and manifest-based
+//!   events, nested struct fields (flattened with dot-notation names), basic
+//!   scalar and string property types, 32-bit and 64-bit event payloads.
 //!
-//! - **Not yet supported** (future work): manifest-based event decoding,
-//!   map / enum value resolution, array-typed properties, and properties
-//!   whose length or count is given by another property.
+//! - **Not supported**: classic (MOF/WBEM) and WPP events (fast-rejected /
+//!   returned as `Unsupported`).
+//!
+//! - **Not yet supported** (future work): map / enum value resolution,
+//!   array-typed properties, and properties whose length or count is given
+//!   by another property.
+//!
+//! ## "Manifest" here means *OS-registered*, not EventSource in-band
+//!
+//! The manifest path resolves schemas exclusively through
+//! `TdhGetEventInformation`, which consults manifests registered with the OS
+//! (`HKLM\...\WINEVT\Publishers`, the `wevtutil im` model — e.g.
+//! `Microsoft-Windows-Kernel-*` and the .NET runtime providers).  It does
+//! **not** support .NET `EventSource` providers running in their default
+//! manifest mode, where the manifest is *not* OS-registered but is emitted
+//! **in-band** as special chunked "manifest events" (`EVENT_DESCRIPTOR.Id ==
+//! 0xFFFE`, `Opcode == 0xFE`).  For those providers `TdhGetEventInformation`
+//! returns `ERROR_NOT_FOUND`, so `decode()` returns
+//! [`TdhDecodeError::NotFound`].  Decoding them requires capturing and
+//! reassembling the in-band manifest chunks and parsing the XML — a separate
+//! subsystem tracked as future work.  Custom `EventSource` providers that use
+//! *self-describing* (TraceLogging) mode are decoded via the TraceLogging
+//! path and are unaffected.
+//!
+//! Manifest decoding requires the provider's manifest to be registered on
+//! the *decoding* machine; if it isn't, `TdhGetEventInformation` returns
+//! `ERROR_NOT_FOUND` and `decode()` returns [`TdhDecodeError::NotFound`].
+//! Negative results are not cached, so a manifest registered after the
+//! decoder starts is picked up on the next matching event.
 
 use super::abi::{EVENT_RECORD, EventRecordExt};
+use crate::Guid;
 use crate::event::{EventData, EventField, EventFormat, LocationType};
 
 use std::collections::HashMap;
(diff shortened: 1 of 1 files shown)


one_collect/src/etw/tdh.rs
```diff
@@ -323,6 +459,107 @@ impl TdhDecoder {
...
+    ///    rather than on every event.  `NotFound` (unregistered manifest) is
+    ///    *not* negatively cached, since it can flip once a manifest registers.
+    fn decode_manifest(
+        &mut self,
+        record: &EVENT_RECORD,
+        is_32bit: bool,
+    ) -> Result<usize, TdhDecodeError> {
+        // Guard 1: classic (MOF/WBEM) events are not manifest events.
+        if (record.EventHeader.Flags & EVENT_HEADER_FLAG_CLASSIC_HEADER) != 0 {
+            return Err(TdhDecodeError::NotFound);
```
utpilla: Shouldn't this be `NotSupported` instead of `NotFound`?

one_collect/src/etw/tdh.rs
```diff
@@ -103,9 +152,19 @@ const PROPERTY_PARAM_COUNT: i32  = PropertyParamCount;
 /// Errors that can occur during TDH-based schema decoding.
 #[derive(Debug)]
 pub enum TdhDecodeError {
```
utpilla, line 155: Let's mark this enum `#[non_exhaustive]`.

one_collect/src/etw/tdh.rs
```diff
@@ -26,18 +34,45 @@
...
+//!   by another property.
+//!
+//! ## "Manifest" here means *OS-registered*, not EventSource in-band
 //!
-//! - **Not yet supported** (future work): manifest-based event decoding,
-//!   map / enum value resolution, array-typed properties, and properties
-//!   whose length or count is given by another property.
+//! The manifest path resolves schemas exclusively through
+//! `TdhGetEventInformation`, which consults manifests registered with the OS
+//! (`HKLM\...\WINEVT\Publishers`, the `wevtutil im` model — e.g.
```
utpilla: nit: Remove em dashes from the doc comments.

one_collect/src/etw/tdh.rs
```diff
@@ -633,13 +894,24 @@ const _: () = assert!(
     "TRACE_EVENT_INFO_0 union must remain 4 bytes",
 );
```
utpilla, line 987: Let's strengthen the `TRACE_EVENT_INFO_0` union guard. The unsafe read of `Anonymous1.EventNameOffset `is sound only because the union is layout-compatible with `u32`. The current `const` guard pins just the size via a magic `== 4`. Suggest pinning both size and alignment to `u32`'s directly, self-documenting and adds alignment coverage the old check missed:


```suggestion
//
// Pin the union's layout to `u32`'s (both size *and* alignment) so the
// `tei.Anonymous1.EventNameOffset` read stays sound: a future `windows-sys`
// bump that grows the union or changes its alignment fails the build here
// rather than silently mis-reading the field.
const _: () = assert!(
    std::mem::size_of::<
        windows_sys::Win32::System::Diagnostics::Etw::TRACE_EVENT_INFO_0
    >() == std::mem::size_of::<u32>(),
    "TRACE_EVENT_INFO_0 union must stay the size of a u32",
);
const _: () = assert!(
    std::mem::align_of::<
        windows_sys::Win32::System::Diagnostics::Etw::TRACE_EVENT_INFO_0
    >() == std::mem::align_of::<u32>(),
    "TRACE_EVENT_INFO_0 union must stay u32-aligned",
);
```

one_collect/src/etw/tdh.rs
```diff
@@ -543,19 +780,43 @@ fn walk_properties(
...
+    // SAFETY: the cast to `*const TRACE_EVENT_INFO` requires the buffer to be
+    // suitably aligned.  Callers pass the `AlignedTeiBuf` backing store, which
+    // is `Vec<u64>`-backed and therefore 8-byte aligned (>= the alignment of
+    // `TRACE_EVENT_INFO`).  The debug assert guards against a future caller
+    // handing in an unaligned slice.
+    debug_assert_eq!(
+        tei_buf.as_ptr() as usize % std::mem::align_of::<TRACE_EVENT_INFO>(),
+        0,
+        "tei_buf must be aligned for TRACE_EVENT_INFO"
+    );
```
utpilla: `tei_buf` is always `AlignedTeiBuf::as_bytes()`, whose pointer is a `Vec<u64>` data pointer (8-byte aligned by construction), so this `debug_assert` on a runtime pointer can become a compile-time const assertion placed next to `AlignedTeiBuf` (it also covers the same cast in `build_cached_schema`), then the `debug_assert` can be dropped.

Don't hardcode `u64`, or it silently breaks if the backing store changes. Tie it to the element type:

```rust
type TeiBufElem = u64;
struct AlignedTeiBuf { storage: Vec<TeiBufElem>, len: usize }

const _: () = assert!(
    std::mem::align_of::<TeiBufElem>() >= std::mem::align_of::<TRACE_EVENT_INFO>(),
    "AlignedTeiBuf backing element must satisfy TRACE_EVENT_INFO alignment",
);
```

one_collect/src/etw/tdh.rs
```diff
@@ -211,24 +351,26 @@ impl SchemaCache {
...
-    /// The TraceLogging event name, or `None` if the schema has no name.
+    /// The resolved event name, or `None` if the schema has no name.
     ///
-    /// This is the primary identity for TraceLogging events (as opposed
-    /// to the event ID, which is often 0).  Consumers can use this for
-    /// OTEL log record naming without a second cache probe.
+    /// For TraceLogging this is the event name (the primary identity, as
+    /// the event ID is often 0); for manifest events it is the task name.
+    /// Consumers can use this for OTEL log record naming without a second
+    /// cache probe.
```
utpilla, line 367: For manifest events `event_name` is the Task name, which is not unique per event: a single manifest Task can back several events distinguished by `Id`/`Opcode` (e.g. `Start`/`Stop`/`Info` under one Task all report the same Task name). That differs from `TraceLogging`, where `event_name` is the unique event name. Worth a doc note on `TdhDecodedEvent::event_name` so identity-sensitive consumers (dedup, routing, record identity) use `schema_id` (or Id/Opcode) rather than treating the name as unique. Suggested addition:

```rust
/// Note: for manifest events this is the Task name, which is not unique
/// per event (multiple events can share a Task). Use `schema_id` if you
/// need a per-schema identity.
```

swashtek:
@utpilla Thank you for feedback, I've addressed your comments in [19ae1fe](https://github.com/microsoft/one-collect/pull/313/commits/19ae1fe3d517cc896abe3068c9cbb9962f70c8fb)

utpilla (approved):

one_collect/src/etw/tdh.rs
```diff
@@ -143,50 +204,112 @@ struct CachedSchema {
...
+///
+/// Note: the derived `Hash` delegates to [`Guid`]'s `Hash` impl, which hashes
+/// only the first three GUID fields (`data1`/`data2`/`data3`), not the
+/// trailing `data4` bytes.  Equality is still full (derived `Eq` compares all
+/// fields), so correctness is unaffected, but two provider GUIDs differing
+/// only in their last 8 bytes share a hash bucket.  Provider GUIDs vary in
+/// their leading fields in practice, so this is benign; it is called out here
+/// so a future change to `Guid`'s `Hash` isn't made without weighing this key.
+#[derive(Copy, Clone, PartialEq, Eq, Hash)]
+struct ManifestKey {
```
brianrob:
```suggestion
struct ManifestEventKey {
```

one_collect/src/etw/tdh.rs
```diff
@@ -271,36 +421,29 @@ impl TdhDecoder {
...
-                    is_32bit,
-                    "TDH schema cache miss — new schema cached"
-                );
-                self.cache.insert(schema_tl_bytes.to_vec(), is_32bit, schema)
-            }
+
+        // Source selection: inline SCHEMA_TL → TraceLogging; otherwise the
+        // event is manifest-based.  `usize` is `Copy`, so the mutable borrow
+        // of `self` in each helper ends before `self.cache.get(idx)` below.
+        let idx = match find_schema_tl(record) {
```
brianrob: It would be nice to centralize the caching behavior in a single place instead of across the different decode methods.  By the time we call `decode_tracelogging` or `decode_manifest` it would be good to be actually attempting a real decode operation.  This will simplify the flow later on when we're dealing with EventSource manifest support where we're going to be injecting entries directly into the cache and not going through the registered manifest path.

one_collect/src/etw/tdh.rs
```diff
@@ -143,50 +204,112 @@ struct CachedSchema {
...
         }
     }
 
-    /// Returns the index of a cached schema, or `None` on a miss.
-    /// The key is hashed exactly once.
-    fn index_of(&self, key: &[u8], is_32bit: bool) -> Option<usize> {
-        let map = if is_32bit { &self.cache_32 } else { &self.cache_64 };
+    /// Returns the index of a cached TraceLogging schema, or `None` on a
+    /// miss.  The key is hashed exactly once.
+    fn index_of_tl(&self, key: &[u8], is_32bit: bool) -> Option<usize> {
```
brianrob, line 280: Question - should we consider a scheme that preserves `index_of` and `insert` instead of having TL and manifest specific methods?  In this scheme, the methods would take metadata as a single parameter, perhaps an enum that would contain either the TL key info or the `ManifestKey` as well as `is_32bit` and `event_type` which could be `TraceLogging` or `Manifest`.  This would allow for selection of the cache in a single method instead of having to pick the right one.  This would simplify the API design.

brianrob (commented):
Thanks for putting this together.  A few questions / comments below.

one_collect/src/etw/tdh.rs
swashtek: Good idea. The enum is actually already here as `SchemaSource`, and `lookup_cached` / `insert_cached` match on it today, so we're most of the way to what you're describing. 
The remaining question is whether to push that match down into `SchemaCache` itself with a unified `index_of(meta)` / `insert(meta, schema)`.

This relocates the branch rather than removing it. The two positive maps have different key types (`HashMap<Vec<u8>, _>` for TL vs `HashMap<ManifestEventKey, _>` for manifest), so a unified `index_of` still needs a 2-arm match inside to select the map pair, the "pick the right cache" step just moves into the cache instead of disappearing. 

Additionally, lookup wants to borrow the TL key (`&[u8]`) while insert must own it (`Vec<u8>`), and the negative cache (`is_manifest_unsupported`) is manifest-only with no TL analog, so it stays separate regardless.

If we do fold it, I'd reuse the existing `SchemaSource` variant as the discriminant (rather than adding a parallel `event_type` field, which could drift out of sync with the variant) and keep `is_32bit` as its own parameter since it's orthogonal to the key.

Since this is purely an internal API-shape change with no behavior difference, I'd suggest we land it as a small follow-up so this PR stays scoped to the manifest/caching functionality, happy to open that up right after. 
Let me know your thoughts.

one_collect/src/etw/tdh.rs
swashtek: Addressed in 1370e9c

one_collect/src/etw/tdh.rs
swashtek: Addressed in 1370e9c

one_collect/src/etw/tdh.rs
brianrob: Sounds great.  Your latest commit looks good too.  Happy to consider the API shape change in a subsequent PR.

brianrob (approved):

PR #314, opened 2026-07-21
Integration test for TDH decoder to support manifest-based events

swashtek:
Integration test for TDH decoder to support manifest-based events
Issue: https://github.com/microsoft/one-collect/issues/310


one_collect/tests/etw_tdh_integration.rs
```diff
@@ -982,3 +1005,97 @@ fn tdh_decodes_tracelogging_static_nested_struct() {
...
+/// (classic-header reject → `TdhGetEventInformation` → `DecodingSource ==
+/// DecodingSourceXMLFile` → schema build) succeeded.  The assertions
+/// additionally prove:
+///
+/// * the manifest resolved to a real field layout (non-empty fields), and
+/// * the `TaskNameOffset` event-name path works (a `ProcessStart` /
+///   `ProcessStop` Task name is surfaced).
+#[ignore]
+#[test]
+fn tdh_decodes_manifest_kernel_process_events() {
```
utpilla, line 1093: We could stop the spawn loop when capture completes. The background spawner here runs for the full TEST_TIMEOUT (30s) unconditionally:

```rust
session.add_started_callback(move |_ctx| {
    std::thread::spawn(move || {
        let deadline = Instant::now() + TEST_TIMEOUT;
        while Instant::now() < deadline {
            let _ = std::process::Command::new("cmd.exe")
                .args(["/c", "exit"])
                ...
                .status();
            std::thread::sleep(Duration::from_millis(25));
        }
    });
});
```


`drive_to_completion` returns as soon as the 10 events are captured (typically well under a second), but this detached thread keeps launching cmd.exe every 25ms (~40/sec, up to ~1200 processes) for the remainder of the 30s. Nothing joins or signals it. It happens to be the last test today, but libtest ordering isn't guaranteed, so under --test-threads=1 any test that ran afterward would inherit that process-spawn noise and CPU load.

We could use a shared stop flag so the spawner exits the instant capture is done, keeping the deadline only as a backstop:

```rust
let stop = Arc::new(AtomicBool::new(false));
let stop_for_thread = stop.clone();

session.add_started_callback(move |_ctx| {
    let stop = stop_for_thread.clone();
    std::thread::spawn(move || {
        let deadline = Instant::now() + TEST_TIMEOUT;
        while !stop.load(Ordering::Relaxed) && Instant::now() < deadline {
            let _ = std::process::Command::new("cmd.exe")
                .args(["/c", "exit"])
                .stdin(std::process::Stdio::null())
                .stdout(std::process::Stdio::null())
                .stderr(std::process::Stdio::null())
                .status();
            std::thread::sleep(Duration::from_millis(25));
        }
    });
});

let captured = drive_to_completion(session, "one_collect_tdh_manifest", &captured, counter, 10);
stop.store(true, Ordering::Relaxed);
```

This preserves the guaranteed-observation property (spawning still only starts after EnableTraceEx2) while spawning a handful of processes instead of hundreds, and removes the cross-test interference risk.

utpilla (commented):
Left some suggestions but looks good overall.

one_collect/tests/etw_tdh_integration.rs
utpilla, line 1: We could consider adding this to a separate file that's dedicated for manifest based events just to keep things simple and organized.

one_collect/tests/etw_tdh_integration.rs
swashtek: Addressed in [7d56b74](https://github.com/microsoft/one-collect/pull/314/commits/7d56b7457915dce0fc16e1fff22ac5543d59868c)

one_collect/tests/etw_tdh_integration.rs
swashtek: The manifest test reuses the same harness (`make_capture_sink`, `build_capturing_session`, `drive_to_completion`) and the `--test-threads=1` global-resource contract as the TraceLogging tests, they're really two decode paths of the same TDH decoder. 

Splitting would mean extracting those helpers into a shared `tests/common` module (or duplicating them), which adds more moving parts than it removes and above private helper `fn`s need to change to `pub(crate)`. 
Right now, it's all private with none of that overhead.   

I suggest keeping them together for now; if the file grows further, we can extract a `common` module and split cleanly at that point. Wdyt?

utpilla: Makes sense.

one_collect/tests/etw_tdh_integration.rs
```diff
@@ -982,3 +1005,107 @@ fn tdh_decodes_tracelogging_static_nested_struct() {
...
+    });
+
+    // Collect a batch so the sample is very likely to include Task-bearing
+    // ProcessStart / ProcessStop events (rather than only Task-less ones).
+    let captured = drive_to_completion(
+        session,
+        "one_collect_tdh_manifest",
+        &captured,
+        counter,
+        10,
```
brianrob: From a reliability perspective, I'm thinking that you should probably watch for the events that you want to validate - even if they don't parse properly, just knowing that they're there, rather than waiting for 10 events seems like it would be more reliable.

swashtek: I agree with this feedback, I'll switch from fixed event count to event detection for determinism.

swashtek: Addressed in [f770f5a](https://github.com/microsoft/one-collect/pull/314/commits/f770f5a034734c78a760e5364c133bae56e4f780)

brianrob (approved):
LGTM.  Thanks!

PR #317, opened 2026-07-24
ETW: Add APIs for resolving ETW provider names to provider GUIDs

### LINKED ISSUE #316: ETW: Provider name to GUID resolution via `One-Collect` APIs
## Summary

Add public ETW helpers that resolve provider names to the GUIDs needed by `EtwSession::enable_provider`.

Today, `enable_provider` only accepts a `Guid`, so any consumer starting from a provider name has to implement its own name-to-GUID resolution. That logic differs by provider model:

- **Manifest/classic MOF providers** must be looked up from the OS provider database.
- **TraceLogging / TraceLoggingDynamic providers** use the EventSource name-hash convention and can be derived directly from the provider name.

This issue proposes exposing both resolution paths from `one_collect` so consumers can reuse a single implementation instead of duplicating TDH enumeration code and EventSource hashing logic.

## Proposed API

```rust
#[derive(Clone, Copy)]
pub struct RegisteredProvider {
    pub guid: Guid,
    pub schema_source: u32,
}

pub fn registered_providers() -> anyhow::Result<HashMap<String, RegisteredProvider>>;

impl Guid {
    pub fn from_eventsource_name(name: &str) -> Guid;
}

pub fn guid_from_tracelogging_name(provider_name: &str) -> anyhow::Result<Guid>;
```

### In scope

- Add `registered_providers()` in `one_collect::etw` to enumerate OS-registered ETW providers via `TdhEnumerateProviders` and return a case-insensitive `name -> RegisteredProvider` map.
- Add a public `RegisteredProvider` type carrying the resolved control `guid` and `schema_source`.
- Expose the EventSource/TraceLogging name-hash convention as `Guid::from_eventsource_name`.
- Add a Windows-facing `guid_from_tracelogging_name` wrapper in `one_collect::etw`.
- Refactor existing internal EventSource hashing code to delegate to the new shared implementation.

### Out of scope

- A policy layer that decides when to use registered lookup vs. TraceLogging name-hash.
- Global caching across calls.
- Adding `EtwSession::enable_provider_by_name`.
- Explicit WPP support.

## Design notes

- Place this surface in `etw`, next to `EtwSession::enable_provider`, rather than in `etw::tdh`.
- `registered_providers()` should do a single `TdhEnumerateProviders` pass and return an owned map.
- Include both manifest (`SchemaSource == 0`) and classic MOF (`SchemaSource == 1`) providers.
- Normalize map keys to lowercase for case-insensitive lookup.
- For duplicate provider names, keep the first entry.
- `guid_from_tracelogging_name` should cover self-describing TraceLogging / TraceLoggingDynamic names and direct `{GUID}` literals.
- Consumers should try `registered_providers()` first and only fall back to TraceLogging name-hash resolution when the name is absent from the registered map.

## Implementation notes

- Reuse `windows-sys` TDH types/functions already used in the ETW codepath (`TdhEnumerateProviders`, `PROVIDER_ENUMERATION_INFO`, `TRACE_PROVIDER_INFO`).
- Use a properly aligned buffer for provider enumeration; do not reinterpret an under-aligned `Vec<u8>`.
- Preserve the existing `.NET` helper behavior, but route the shared EventSource hash path through `Guid::from_eventsource_name` so there is a single copy of the namespace/encoding logic.

## Motivation

This is primarily needed by downstream ETW consumers that are configured by provider name rather than GUID. In particular, the `otel-arrow` ETW receiver currently carries both:

- its own `TdhEnumerateProviders` FFI for registered provider lookup, and
- a duplicate copy of the EventSource namespace/hash logic.

Publishing this API from `one_collect` would let downstream consumers delete that duplicated logic and keep only their configuration policy.

## Acceptance criteria

- Consumers can resolve manifest/classic MOF provider names through `registered_providers()`.
- Consumers can resolve TraceLogging / TraceLoggingDynamic names through `Guid::from_eventsource_name` / `guid_from_tracelogging_name`.
- Internal duplicated EventSource hash logic is consolidated behind the new public API.
- The API is released in a published `one_collect` version so downstream consumers can adopt it.

swashtek:
## Goal:
- Add public ETW provider enumeration through `for_each_registered_provider`
- Add the public `guid_from_provider_name` helper for EventSource/TraceLogging provider names

## Scope

This PR exposes the lookup and hashing primitives. Consumers remain responsible for deciding whether to use registered-provider lookup or EventSource/TraceLogging name hashing.

It intentionally does not add:

- A combined name-resolution policy layer.
- `EtwSession::enable_provider_by_name`.
- Global provider caching.
- Explicit WPP support.

## Issue: 
Fixes https://github.com/microsoft/one-collect/issues/316


one_collect/src/etw/providers.rs
```diff
@@ -0,0 +1,331 @@
...
+};
+
+/// A provider found in the OS-registered provider database.
+#[derive(Clone, Copy)]
+pub struct RegisteredProvider {
+    /// Control GUID of the registered provider.
+    pub guid: Guid,
+    /// TDH schema source: `0` = XML manifest, `1` = classic WMI/MOF.  Retained
+    /// so callers can report which decoding model a matched provider uses.
+    pub schema_source: u32,
```
brianrob: I'd recommend that this be an enum to make it's valid values clearer.

one_collect/src/etw/providers.rs
```diff
@@ -0,0 +1,331 @@
...
+            continue;
+        };
+
+        let provider = RegisteredProvider {
+            guid: win32_guid_to_guid(&info.ProviderGuid),
+            schema_source: info.SchemaSource,
+        };
+
+        // Use Unicode lowercase for case-insensitive matching semantics.
+        if f(&provider_name.to_lowercase(), provider).is_break() {
```
brianrob: I'm wondering if we should be calling `to_lower()` here or if we should let the user do a case-insensitive comparison if they want to.

one_collect/src/etw/providers.rs
```diff
@@ -0,0 +1,331 @@
...
+        anyhow::bail!("TdhEnumerateProviders (size probe) failed with Win32 error {status}");
+    }
+    if (buffer_size as usize) < std::mem::size_of::<PROVIDER_ENUMERATION_INFO>() {
+        // Empty or header-less result: nothing to enumerate.
+        return Ok(());
+    }
+
+    // Fill call with a bounded retry: providers can be registered between
+    // probe and fill, causing one transient `ERROR_INSUFFICIENT_BUFFER`.
+    let mut retries_remaining = 1u8;
```
brianrob: I'd recommend making this number a bit higher to avoid cases where providers get registered concurrently with the call.  Maybe 5?

one_collect/src/etw/providers.rs
```diff
@@ -0,0 +1,331 @@
...
+                    out.as_mut_ptr().add(off) as *mut TRACE_PROVIDER_INFO,
+                    info,
+                )
+            };
+        }
+
+        out
+    }
+
+    #[test]
```
brianrob, line 309: Would be good to add a test here for the "happy path" that iterates through multiple providers and asserts over their values.

one_collect/src/etw/providers.rs
brianrob: Could this be a typed enum instead of exposing TDH's magic numbers in the public API? I'd suggest `Manifest`, `Wmi`, and `Unknown(u32)` variants, converting from `info.SchemaSource` at the Win32 boundary so unknown values remain forward-compatible.

brianrob: Could this be a typed enum instead of exposing TDH's magic numbers in the public API? I'd suggest `Manifest`, `Wmi`, and `Unknown(u32)` variants, converting from `info.SchemaSource` at the Win32 boundary so unknown values remain forward-compatible.

one_collect/src/lib.rs
```diff
@@ -72,6 +72,42 @@ impl Guid {
...
+///
+/// Unlike manifest / classic providers (whose GUID must be looked up in the
+/// OS provider database), a self-describing provider's GUID is a deterministic
+/// hash of its name: the name is upper-cased, encoded as UTF-16BE, and
+/// v5-hashed over the fixed 16-byte `EventSource` namespace.
+///
+/// This is a pure name-hash function. It does not parse `{GUID}` literals;
+/// caller policy decides whether input should be treated as a literal GUID or
+/// hashed as a provider name.
+pub fn guid_from_provider_name_hash(name: &str) -> Guid {
```
brianrob: Consider the name `guid_from_provider_name`.

brianrob (commented):
Overall looks good - a few comments/questions.

one_collect/src/etw/providers.rs
swashtek: Sure, I can make it 5, it is still tolerant enough for few concurrent registrations between. 
If churn is high enough, it will fail fast and won't loop. 
Addressed in [b489684](https://github.com/microsoft/one-collect/pull/317/commits/b48968451bf4e32dce543d43d601c00b64a71185)

one_collect/src/etw/providers.rs
swashtek: Makes sense, addressed in [b489684](https://github.com/microsoft/one-collect/pull/317/commits/b48968451bf4e32dce543d43d601c00b64a71185)

one_collect/src/etw/providers.rs
swashtek: Addressed in [b489684](https://github.com/microsoft/one-collect/pull/317/commits/b48968451bf4e32dce543d43d601c00b64a71185)

one_collect/src/etw/providers.rs
swashtek: I agree. I removed the internal lowercasing and now pass provider names through exactly as registered by the OS, so callers can choose their own comparison semantics.

one_collect/src/etw/providers.rs
swashtek: Replaced `[u32]` with `ProviderSchemaSource { Manifest, Wmi, Unknown(u32) }` and convert from `info.SchemaSource` at enumeration time.

swashtek: Agreed, `schema_source` is now a typed enum with `Manifest`, `Wmi` and `Unknown(u32)`, and conversion happens at the TDH boundary when reading `info.SchemaSource` so unknown values stay forward-compatible.

swashtek:
> Overall looks good - a few comments/questions.

@brianrob Thank you for the feedback, I've addressed it. Could you please give another pass at it? 
Thanks again.

one_collect/src/etw/mod.rs
```diff
@@ -15,8 +15,11 @@ use crate::Guid;
 #[allow(dead_code)]
 mod abi;
 mod events;
+pub mod providers;
```
utpilla: The module is public and its items are also re-exported below, so there are two public paths to the same API (now three items). Make the module private and keep only the re-export.

one_collect/src/etw/providers.rs
```diff
@@ -0,0 +1,408 @@
...
+                );
+            }
+            // Parse only the byte count returned by TDH, not the full
+            // allocation length.
+            // SAFETY: `buffer` owns `alloc_byte_len` valid bytes; `used_byte_len`
+            // is range-checked above.
+            let used = unsafe {
+                std::slice::from_raw_parts(buffer.as_ptr() as *const u8, used_byte_len)
+            };
+            break used.to_vec();
```
utpilla: minor: The success arm copies the whole provider table into a fresh Vec<u8> (break used.to_vec();) just to parse it after the loop. You can parse in place and return, dropping final_bytes and the trailing call:

```rust
let used = unsafe { std::slice::from_raw_parts(buffer.as_ptr() as *const u8, used_byte_len) };
return for_each_provider_in_buffer(used, &mut f);
```

one_collect/src/etw/providers.rs
```diff
@@ -0,0 +1,408 @@
...
+//! `TdhEnumerateProviders` API to visit every provider registered in the
+//! system provider database, invoking a caller-supplied closure with each
+//! provider's name and [`RegisteredProvider`] details.
+//!
+//! The database is enumerated in a single pass, so the closure sees every
+//! provider from one `TdhEnumerateProviders` call. The caller decides what to
+//! keep (e.g. matching a configured set of names) and can stop early by
+//! returning [`ControlFlow::Break`], so no full map is ever materialized.
+//! Both manifest-based providers (`SchemaSource == 0`) and classic MOF
+//! providers (`SchemaSource == 1`) are visited.
```
utpilla: We can be clearer here.

```suggestion
//! Every registered provider is surfaced regardless of schema source;
//! unrecognized sources are reported as `ProviderSchemaSource::Unknown`.
```

one_collect/src/etw/providers.rs
```diff
@@ -0,0 +1,408 @@
...
+    Manifest,
+    /// Provider schema uses classic WMI/MOF (`SchemaSource == 1`).
+    Wmi,
+    /// Any schema source value not currently recognized by this crate.
+    Unknown(u32),
+}
+
+impl ProviderSchemaSource {
+    /// Return the raw TDH `SchemaSource` value.
+    pub fn as_raw(self) -> u32 {
```
utpilla:
```suggestion
    pub const fn as_raw(self) -> u32 {
```

one_collect/src/etw/providers.rs
```diff
@@ -0,0 +1,408 @@
...
+    pub guid: Guid,
+    /// TDH schema source for the registered provider.
+    pub schema_source: ProviderSchemaSource,
+}
+
+/// Convert a Win32 `GUID` (from `windows-sys`) into a [`Guid`].
+///
+/// Both types are `#[repr(C)]` with identical `data1/2/3/4` fields, so this is
+/// a straight field copy.
+fn win32_guid_to_guid(g: &windows_sys::core::GUID) -> Guid {
```
utpilla:
```suggestion
const fn win32_guid_to_guid(g: &windows_sys::core::GUID) -> Guid {
```

one_collect/src/etw/providers.rs
```diff
@@ -0,0 +1,408 @@
...
+        anyhow::bail!("TdhEnumerateProviders (fill) failed with Win32 error {status}");
+    };
+
+    for_each_provider_in_buffer(&final_bytes, &mut f)?;
+
+    Ok(())
+}
+
+#[allow(unsafe_code)]
+fn for_each_provider_in_buffer(
```
utpilla, line 169: Consider adding two `const` assertions at the top of `for_each_provider_in_buffer`, right before the `unsafe` reads. The parser already depends on these layout facts; asserting them turns a future `windows-sys` change into a build error instead of runtime corruption:

- The entry offset uses `saturating_sub` between `size_of::<PROVIDER_ENUMERATION_INFO>()` and `size_of::<TRACE_PROVIDER_INFO>()`. If the header ever became smaller than one entry, that would silently go to 0 and every provider would be read from the wrong offset.
- `max_entries` divides by `size_of::<TRACE_PROVIDER_INFO>()`, so it must not be zero.

```rust
// Layout facts the unsafe reads below rely on. Alignment is not checked
// because every struct read uses `read_unaligned`.
const _: () = assert!(
    std::mem::size_of::<PROVIDER_ENUMERATION_INFO>()
        >= std::mem::size_of::<TRACE_PROVIDER_INFO>()
);
const _: () = assert!(std::mem::size_of::<TRACE_PROVIDER_INFO>() > 0);
```

one_collect/src/etw/providers.rs
```diff
@@ -0,0 +1,408 @@
...
+}
+
+/// Read a null-terminated UTF-16 (little-endian) string starting at
+/// `byte_offset` within `bytes`, stopping at the terminator or the end of the
+/// buffer. Returns `None` if the offset leaves no room for even one code unit.
+fn read_utf16z(bytes: &[u8], byte_offset: usize) -> Option<String> {
+    if byte_offset + 1 >= bytes.len() {
+        return None;
+    }
+    let mut units = Vec::new();
```
utpilla: We could avoid this allocation by the decoding directly to `String`.

```rust
fn read_utf16z(bytes: &[u8], byte_offset: usize) -> Option<String> {
    if byte_offset + 1 >= bytes.len() {
        return None;
    }
    let units = bytes[byte_offset..]
        .chunks_exact(2)
        .map(|c| u16::from_le_bytes([c[0], c[1]]))
        .take_while(|&u| u != 0);
    Some(
        char::decode_utf16(units)
            .map(|r| r.unwrap_or(char::REPLACEMENT_CHARACTER))
            .collect(),
    )
}
```

This keeps the same behavior: `chunks_exact(2)` ignores a trailing odd byte (like the current `i + 1 < len` loop), and mapping decode errors to `REPLACEMENT_CHARACTER` matches the existing lossy conversion. Just one allocation instead of two, and no manual index arithmetic.

utpilla (approved):
Left some non-blocking suggestions.

one_collect/src/etw/providers.rs
swashtek: Good suggestion, updated. `win32_guid_to_guid` is now a `const fn`; behavior is unchanged.

one_collect/src/etw/providers.rs
swashtek: Great point, added.

one_collect/src/etw/providers.rs
swashtek: Updated, `read_utf16z` now decodes directly into `String` using `decode_utf16`, preserving prior lossy behavior and odd-byte handling while removing the intermediate `Vec<u16>` allocation.

swashtek:
> Left some non-blocking suggestions.

Thank you for suggestions, I've addressed them.

brianrob (approved):
LGTM.  Thanks @swashtek.

@utpilla, please take a look at the changes associated with your comments.  Once you approve, I'll merge.

utpilla (approved):

PR #318, opened 2026-07-24
Bound Go symbol reader memory and add ELF comparison benchmarks

brianrob:
## Summary

- change Go symbol resolution to iterate through \`.gopclntab\` instead of loading the entire table
- bound reader memory to 196 KiB, independent of binary size
- simplify decoding around one shared cursor with two fixed read windows
- add reproducible Go fixtures and benchmarks comparing Go and ELF resolution

## Performance

Using matched fixtures with 1,273 symbols:

| Benchmark | ELF | Go | Result |
|---|---:|---:|---:|
| Reader setup | 29.39 µs | 46.00 µs | Go setup is 1.57× slower |
| Full lifecycle | 2.373 ms | 153.82 µs | Go is 15.43× faster |
| Mapping resolution | 2.656 ms | 341.58 µs | Go is 7.78× faster |
| ELF+Go union mapping | 2.656 ms ELF-only | 2.877 ms union | 8.32% overhead |
| Production resolution | 3.128 ms unstripped ELF+Go | 592.78 µs stripped Go | Stripped Go is 5.28× faster |


one_collect/src/ruwind/elf.rs
```diff
@@ -717,414 +717,596 @@ pub fn read_debug_link<'a>(
...
+        seg_offset: u64,
+        seg_filesz: u64,
+        seg_vaddr: u64,
+        text_vaddr: Option<u64>,
+        scan_chunk_size: usize,
+    ) -> Option<crate::ruwind::go_pclntab::GoPclnTabLocation> {
+        const SCAN_OVERLAP: usize = 7;
+
+        let seg_end = seg_offset.checked_add(seg_filesz)?;
+        let mut bytes = vec![0u8; scan_chunk_size.checked_add(SCAN_OVERLAP)?];
```
beaubelgrave: This is done per-executable segment, this can be lifted up to callers and re-used. For binaries that have lots of segments this might be an issue.

one_collect/src/ruwind/go_pclntab.rs
```diff
@@ -16,156 +16,572 @@
...
+    data: Box<[u8]>,
+    base: u64,
+    align: bool,
+    start: u64,
+    len: usize,
+    #[cfg(test)]
+    refills: usize,
+}
+
+impl ReadWindow {
```
beaubelgrave, line 127: Do we think ELF would benefit from these as well? Seems useful to use in general.

beaubelgrave (approved):
Non-blocking comments added, seems like a useful pattern we can extend to ELF as well. A few opportunities to reduce allocations called out.

one_collect/src/ruwind/go_pclntab.rs
brianrob: Filed #319 to track this.  Could be interesting.

one_collect/src/ruwind/elf.rs
brianrob: Good call.  I've committed a change to do this.

beaubelgrave (approved):
LGTM

PR #320, opened 2026-08-10
Support manifest counted strings in the TDH decoder

utpilla:
## Summary

Adds decoding support for manifest-range counted strings (`TDH_INTYPE_MANIFEST_COUNTEDSTRING` = 22 and `TDH_INTYPE_MANIFEST_COUNTEDANSISTRING` = 23) to the ETW TDH decoder, and adds an end-to-end integration test that compiles, registers, emits, and decodes a real manifest covering every `inType` the decoder supports.

## Motivation
Providers that switch a manifest field from `UnicodeString` to `CountedUnicodeString` (a common hardening step, since counted strings are immune to the NUL-truncation injection where attacker content after a NUL is parsed as later columns) are compiled by `mc.exe` into `TDH_INTYPE_MANIFEST_COUNTEDSTRING` (22). The decoder previously only handled the WBEM/WMI-range counted strings (`TDH_INTYPE_COUNTEDSTRING` = 300, `TDH_INTYPE_COUNTEDANSISTRING` = 301), so intype 22/23 fell through to the `"unsupported"` catch-all (`Static`, size 0). That silently consumed 0 bytes and misaligned every following field, rather than erroring. The two families share an identical on-wire layout (a 2-byte byte-count length prefix followed by the content), so the same decode path handles both.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -118,8 +118,17 @@ jobs:
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
       - name: Run etw_tdh_integration
+        # The manifest-decode test (`tdh_decodes_compiled_manifest_all_types`)
+        # compiles a WES manifest at runtime, so mc.exe / rc.exe / link.exe
+        # must be on PATH.  Initialise the VS dev environment via vcvars64
+        # (located with vswhere) rather than a third-party action.
+        # wevtutil.exe is already on PATH and the runner is `runneradmin`.
         working-directory: one_collect
-        run: cargo test --locked --test etw_tdh_integration -- --ignored --nocapture --test-threads=1
+        shell: cmd
+        run: |
+          for /f "usebackq delims=" %%i in (`"%ProgramFiles(x86)%\Microsoft Visual Studio\Installer\vswhere.exe" -latest -property installationPath`) do set "VSPATH=%%i"
+          call "%VSPATH%\VC\Auxiliary\Build\vcvars64.bat"
+          cargo test --locked --test etw_tdh_integration -- --ignored --nocapture --test-threads=1
       - name: Run etw_stats_integration
         working-directory: one_collect
         run: cargo test --locked --test etw_stats_integration -- --ignored --nocapture --test-threads=1
--- a/one_collect/tests/etw_tdh_integration.rs
+++ b/one_collect/tests/etw_tdh_integration.rs
@@ -1180,3 +1180,537 @@ fn tdh_decodes_manifest_kernel_process_events() {
     );
 }
 
+// ── Test F: manifest decoding of a compiled-at-test-time manifest ─────
+//
+// Unlike Test E (which drives a live OS provider), this test owns its
+// provider end to end: it compiles the checked-in WES manifest
+// (`test_assets/manifest/one_collect_test.man`) with the Windows SDK
+// (`mc.exe` -> `rc.exe` -> `link.exe`), registers it with the OS
+// (`wevtutil im`), emits one event per template through the live ETW
+// runtime via `EventRegister`/`EventWrite`, and asserts that every inType
+// the decoder supports round-trips through the manifest path.
+//
+// This is the comprehensive counterpart to Test E and is the only test
+// that exercises the manifest counted-string inTypes
+// (`win:CountedUnicodeString` / `win:CountedAnsiString` ->
+// `TDH_INTYPE_MANIFEST_COUNTEDSTRING`/`ANSISTRING`).
+//
+// It requires elevation (for `wevtutil im` and the ETW consumer session)
+// and the Windows SDK + MSVC tools on `PATH`.  When those tools are
+// absent it prints a `SKIP` line and returns `Ok` rather than failing, so
+// a non-elevated / SDK-less dev box still passes `cargo test`.  CI puts
+// the tools on `PATH` (see `msvc-dev-cmd` in the ETW integration job) and
+// runs elevated as `runneradmin`.
+//
+// Scope: x64 only.  The link step hard-codes `/MACHINE:X64` and the
+// payload assumes an 8-byte `win:Pointer`, matching the x64 Windows CI
+// runner; a 32-bit build would need a different machine flag and pointer
+// width.
+
+use std::path::{Path, PathBuf};
+use std::process::Command;
+use windows_sys::Win32::System::Diagnostics::Etw::{
+    EventRegister, EventUnregister, EventWrite,
+    EVENT_DATA_DESCRIPTOR, EVENT_DESCRIPTOR,
+};
+
+/// `u128` form of the test provider GUID, used to build the framework
+/// `Guid` the capturing session subscribes to.
+const TEST_MANIFEST_PROVIDER: u128 = 0x4948EF3B_4F28_4747_9BB9_649008E2EDEF;
+
+/// `windows_sys` form of the same GUID, used by `EventRegister`.
+const PROVIDER_GUID_SYS: windows_sys::core::GUID = windows_sys::core::GUID {
+    data1: 0x4948EF3B,
+    data2: 0x4F28,
+    data3: 0x4747,
+    data4: [0x9B, 0xB9, 0x64, 0x90, 0x08, 0xE2, 0xED, 0xEF],
+};
+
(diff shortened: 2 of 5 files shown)


one_collect/tests/etw_tdh_integration.rs
```diff
@@ -1180,3 +1180,496 @@ fn tdh_decodes_manifest_kernel_process_events() {
...
+                }
+            };
+            let scalars = scalars_payload();
+            let strings = strings_payload();
+            let blob = blob_payload();
+            let deadline = Instant::now() + TEST_TIMEOUT;
+            while !stop.load(Ordering::Relaxed) && Instant::now() < deadline {
+                emitter.write(1, 1, &scalars);
+                emitter.write(2, 2, &strings);
+                emitter.write(3, 3, &blob);
```
swashtek: Would it be worth to capture and report nonzero `EventWrite` status codes instead of silently discarding them?
It will be easier to diagnose than a generic `parse_until` timeout. 
Let me know if it's worth it.

one_collect/tests/etw_tdh_integration.rs
```diff
@@ -1180,3 +1180,496 @@ fn tdh_decodes_manifest_kernel_process_events() {
...
+        &["/nologo", "/fo", "one_collect_test.res", "one_collect_test.rc"],
+        &workdir,
+    )?;
+    // 3. Linker: .res -> resource-only DLL (no code, no entry point).
+    run_tool(
+        "link.exe",
+        &[
+            "/NOLOGO", "/DLL", "/NOENTRY", "/MACHINE:X64",
+            "/OUT:one_collect_test.dll", "one_collect_test.res",
+        ],
```
swashtek: THe test only covers 64 bit which is reasonable given Windows CI, can we explicitly document it ?

one_collect/tests/etw_tdh_integration.rs
```diff
@@ -1180,3 +1180,496 @@ fn tdh_decodes_manifest_kernel_process_events() {
...
+    let dll = workdir.join("one_collect_test.dll");
+    let rf = format!("/rf:{}", dll.display());
+    let mf = format!("/mf:{}", dll.display());
+    run_tool(
+        "wevtutil.exe",
+        &["im", "one_collect_test.man", &rf, &mf],
+        &workdir,
+    )?;
+
+    Ok(RegisteredManifest { man, workdir })
```
swashtek: If any step fails, `workdir` is left behind, is it possible to make cleanup unconditional and explicitly remove `workdir` before returning error?

swashtek (approved):
LGTM, added couple of non-blocking comments. 
Thank you for creating the change.

one_collect/tests/etw_tdh_integration.rs
utpilla: Updated

one_collect/tests/etw_tdh_integration.rs
utpilla: Updated

one_collect/tests/etw_tdh_integration.rs
utpilla: Fixed

utpilla:
@brianrob Could you please review this PR?

.github/workflows/ci.yml
```diff
@@ -116,7 +116,13 @@ jobs:
     name: ETW Integration Tests (Windows-x64)
     runs-on: windows-latest
     steps:
-      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+      - uses: actions/checkout@v3
```
brianrob: Please update this to use the full commit hash instead of the tag.  There is a recent change to switch to this, so you can pull the commit from the latest yml.

.github/workflows/ci.yml
```diff
@@ -116,7 +116,13 @@ jobs:
...
+      - name: Set up MSVC + Windows SDK on PATH
+        # Puts mc.exe / rc.exe / link.exe on PATH for the manifest-decode
+        # integration test (`tdh_decodes_compiled_manifest_all_types`),
+        # which compiles and registers a WES manifest at test time.
+        # wevtutil.exe is already on PATH and the runner is `runneradmin`.
+        uses: ilammy/msvc-dev-cmd@v1
```
brianrob: Do we need a 3rd party action for this? Presumably we should just be able to run the VS dev shell cmd file, right?

one_collect/tests/etw_tdh_integration.rs
```diff
@@ -1180,3 +1180,527 @@ fn tdh_decodes_manifest_kernel_process_events() {
...
+    if !output.status.success() {
+        return Err(format!(
+            "`{name} {}` failed ({}):\nstdout: {}\nstderr: {}",
+            args.join(" "),
+            output.status,
+            String::from_utf8_lossy(&output.stdout),
+            String::from_utf8_lossy(&output.stderr),
+        ));
+    }
+    Ok(())
```
brianrob, line 1317: Seems like it might be worthwhile to log the output of the tool even if it does succeed.

one_collect/tests/etw_tdh_integration.rs
```diff
@@ -1180,3 +1180,527 @@ fn tdh_decodes_manifest_kernel_process_events() {
...
+/// A registered manifest whose `Drop` unregisters it (`wevtutil um`) and
+/// removes the temp build directory, so cleanup happens even on panic.
+struct RegisteredManifest {
+    man: PathBuf,
+    workdir: PathBuf,
+}
+
+impl Drop for RegisteredManifest {
+    fn drop(&mut self) {
+        let _ = Command::new("wevtutil.exe")
```
brianrob: Should this call `run_tool`?

one_collect/tests/etw_tdh_integration.rs
```diff
@@ -1180,3 +1180,527 @@ fn tdh_decodes_manifest_kernel_process_events() {
...
+                "/OUT:one_collect_test.dll", "one_collect_test.res",
+            ],
+            &workdir,
+        )?;
+        // 4. Register (requires admin); point both resource + message files
+        //    at the freshly built DLL by absolute path.
+        let dll = workdir.join("one_collect_test.dll");
+        let rf = format!("/rf:{}", dll.display());
+        let mf = format!("/mf:{}", dll.display());
+        run_tool(
```
brianrob, line 1383: I saw a note in the yml that the test is already running as runneradmin.  Does that mean that the CI is running as admin?

brianrob (commented):
Overall, looks good.  A few questions and comments.

one_collect/tests/etw_tdh_integration.rs
```diff
@@ -1180,3 +1180,527 @@ fn tdh_decodes_manifest_kernel_process_events() {
...
+        Err(e) => {
+            let _ = std::fs::remove_dir_all(&workdir);
+            Err(e)
+        }
+    }
+}
+
+/// Thin RAII wrapper over an ETW provider registration used to emit the
+/// manifest events.
+struct ManifestEmitter {
```
brianrob, line 1405: Don't make any changes here, but thinking forward, we should consider adding this capability to one\_collect.  Ideally, one\_collect can write events to ETW and user\_events on its own so that it can inject its own events into the trace.

.github/workflows/ci.yml
utpilla: Updated.

.github/workflows/ci.yml
utpilla: I have implemented your suggestion.

one_collect/tests/etw_tdh_integration.rs
utpilla: Done

one_collect/tests/etw_tdh_integration.rs
utpilla: Updated.

one_collect/tests/etw_tdh_integration.rs
utpilla: Yes. GitHub-hosted `windows-latest` runners execute as `runneradmin`, which is a member of the local Administrators group, so the job effectively runs elevated. That is what lets the ETW integration tests start a consumer session and (for the new test) run `wevtutil im` to register the compiled manifest.

brianrob (approved):
LGTM.  Thanks @utpilla!

PR #322, opened 2026-08-11
Add cargo-deny check to CI

utpilla:
Adds a cargo-deny check so CI audits the dependency graph for security advisories, disallowed/duplicate crates, license compliance, and untrusted sources. This PR only adds the check and its configuration; it does not change any dependencies.

### DIFF
--- a/.github/workflows/cargo-deny.yml
+++ b/.github/workflows/cargo-deny.yml
@@ -1,0 +1,31 @@
+name: Cargo Deny
+
+permissions:
+  contents: read
+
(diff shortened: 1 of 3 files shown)


brianrob:
Should we close this in favor of #323?

utpilla:
> Should we close this in favor of #323?

@brianrob Yes, I created this PR mainly to show the errors in the cargo deny CI check.

brianrob:
Ah yes, that makes sense.

PR #323, opened 2026-08-11
Resolve cargo-deny findings

utpilla:
## Summary

- Add cargo deny check CI job
- Makes cargo deny check pass for both crate roots by updating a few dependencies, removing an unused one, and scoping/annotating the config for the crates that have no maintained replacement.

Note to reviewers:
- This PR builds on top of #322 (which shows the cargo deny check failures)

### DIFF
--- a/.github/workflows/cargo-deny-advisories.yml
+++ b/.github/workflows/cargo-deny-advisories.yml
@@ -1,0 +1,87 @@
+name: Cargo Deny Advisories
+
+# Scans the full resolved dependency graph for RustSec advisories on a schedule
+# rather than per-PR. This keeps awareness of newly disclosed advisories against
+# existing (already-merged) dependencies without blocking unrelated PRs. The
+# PR-blocking bans/licenses/sources checks live in cargo-deny.yml.
+#
+# On failure this workflow files (or updates) a single tracking issue labeled
+# `cargo-deny-advisories`, assigned to the maintainers, so they get an email and
+# a place to track the fix. A later passing run closes the issue again.
+
+permissions:
+  contents: read
+  issues: write
+
+on:
+  schedule:
+    # Daily at 06:00 UTC.
+    - cron: '0 6 * * *'
+  workflow_dispatch:
+
+env:
+  CARGO_TERM_COLOR: always
+
+jobs:
+  advisories:
+    name: Cargo Deny Advisories
+    runs-on: ubuntu-latest
+    env:
+      GH_TOKEN: ${{ github.token }}
+      ISSUE_LABEL: cargo-deny-advisories
+
+    steps:
+    - uses: actions/checkout@v3
+
+    - name: Install cargo-deny
+      run: cargo install cargo-deny --locked
+
+    # Check both crate roots in a single step so record-trace is always scanned
+    # even when one_collect reports an advisory (separate steps would stop at the
+    # first failure). The combined output is saved for the tracking issue.
+    - name: cargo deny check advisories
+      run: |
+        status=0
+        echo "## one_collect" | tee advisory-report.txt
+        cargo deny --manifest-path one_collect/Cargo.toml --config deny.toml check advisories 2>&1 | tee -a advisory-report.txt || status=1
+        echo "" | tee -a advisory-report.txt
+        echo "## record-trace" | tee -a advisory-report.txt
+        cargo deny --manifest-path record-trace/Cargo.toml --config deny.toml check advisories 2>&1 | tee -a advisory-report.txt || status=1
+        exit $status
+
+    # File a single tracking issue (or comment on the existing open one so we do
+    # not open a new issue every day) when advisories are found.
+    - name: File or update tracking issue on failure
+      if: failure()
+      run: |
+        gh label create "$ISSUE_LABEL" --description "cargo-deny advisory scan failures" --color B60205 --force >/dev/null 2>&1 || true
+        run_url="${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}"
+        {
+          echo "The scheduled \`cargo deny check advisories\` run failed."
+          echo
+          echo "Run: $run_url"
+          echo
+          echo "<details><summary>cargo-deny output</summary>"
+          echo
+          echo '```'
+          cat advisory-report.txt
+          echo '```'
+          echo
+          echo "</details>"
+        } > issue-body.md
+        existing=$(gh issue list --label "$ISSUE_LABEL" --state open --json number --jq '.[0].number')
+        if [ -n "$existing" ]; then
+          gh issue comment "$existing" --body-file issue-body.md
+        else
+          gh issue create --label "$ISSUE_LABEL" --assignee brianrob --assignee beaubelgrave --title "cargo-deny: advisory detected in dependency graph" --body-file issue-body.md
+        fi
+
+    # Self-heal: once a later run passes, close any open tracking issue.
+    - name: Close tracking issue on success
+      if: success()
+      run: |
+        run_url="${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}"
+        for n in $(gh issue list --label "$ISSUE_LABEL" --state open --json number --jq '.[].number'); do
+          gh issue comment "$n" --body "Latest scheduled \`cargo deny check advisories\` run passed ($run_url). Closing."
+          gh issue close "$n"
+        done
--- a/.github/workflows/cargo-deny-advisories.yml
+++ b/.github/workflows/cargo-deny-advisories.yml
@@ -31,7 +31,7 @@ jobs:
       ISSUE_LABEL: cargo-deny-advisories
 
     steps:
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
 
     - name: Install cargo-deny
       run: cargo install cargo-deny --locked
--- a/.github/workflows/cargo-deny.yml
+++ b/.github/workflows/cargo-deny.yml
@@ -23,7 +23,7 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
 
     # Installed from source to avoid taking on a third-party GitHub Action;
     # `--locked` keeps the build reproducible.
--- a/.github/workflows/cargo-deny.yml
+++ b/.github/workflows/cargo-deny.yml
@@ -1,0 +1,31 @@
+name: Cargo Deny
+
+permissions:
+  contents: read
+
+on:
+  workflow_call:
+
+env:
+  CARGO_TERM_COLOR: always
+
+jobs:
+  cargo-deny:
+    name: Cargo Deny
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v3
+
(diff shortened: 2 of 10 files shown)


deny.toml
```diff
@@ -0,0 +1,62 @@
...
+# root (one_collect and record-trace) against this shared config.
+
+[graph]
+# Restrict the audited graph to the platforms one-collect actually ships and
+# tests in CI. This drops target-specific dependencies that never build here
+# (e.g. the wasm-only `instant` pulled in by rhai).
+targets = [
+    "x86_64-unknown-linux-gnu",
+    "x86_64-unknown-linux-musl",
+    "x86_64-pc-windows-msvc",
```
brianrob, line 19: We also build arm64, so we'll need to cover that as well.

deny.toml
```diff
@@ -0,0 +1,62 @@
...
+]
+
+[advisories]
```
brianrob, line 22: My understanding is that this will scan every direct and transitive dependency and fail if there are advisories, even if the PR made no changes to dependencies.  While knowing that there is an advisory is worthwhile, I wouldn't want that to slow down code motion in the repo.  I'd only want this sort of thing to run if it were limited to dependencies added in the current PR.

deny.toml
```diff
@@ -0,0 +1,62 @@
...
+    # smartstring (transitive via rhai) is MPL-2.0, a file-level weak copyleft
+    # that imposes no obligations on the rest of the project. Tied to the
+    # RUSTSEC-2026-0249 ignore above; both go away if rhai drops smartstring.
+    { name = "smartstring", allow = ["MPL-2.0"] },
+]
+
+[bans]
+# Surface duplicate versions of the same crate as a warning.
+multiple-versions = "warn"
+wildcards = "allow"
```
brianrob: I think we should fail wildcards.  We already build with `--locked` and want to make sure that dependencies are explicit.

deny.toml
```diff
@@ -0,0 +1,62 @@
...
+    # removes it. Re-evaluate whenever rhai drops or replaces smartstring.
+    "RUSTSEC-2026-0249",
+]
+
+[licenses]
+# Permissive licenses only. Dual/tri-licensed crates (e.g. `r-efi` is
+# `MIT OR Apache-2.0 OR LGPL-2.1-or-later`) resolve to one of these, so a
+# copyleft option is never selected.
+allow = [
+    "0BSD",
```
brianrob: This allow-list is broader than the currently targeted dependency graphs require. Please remove `0BSD`, `BSD-2-Clause`, and `Zlib`; they appear only as alternatives to MIT or Apache, and leaving them allowed would let future dependencies introduce those licenses without an explicit review.

brianrob (commented):
Some comments below.  Also @beaubelgrave should weigh-in on this one too.

deny.toml
```diff
@@ -0,0 +1,62 @@
...
+#
+# cargo-deny audits the dependency graph for security advisories, disallowed or
+# duplicate crates, license compliance, and untrusted sources. See
+# https://embarkstudios.github.io/cargo-deny/ for the configuration reference.
+#
+# There is no workspace-level Cargo.toml, so CI runs cargo-deny once per crate
+# root (one_collect and record-trace) against this shared config.
+
+[graph]
+# Restrict the audited graph to the platforms one-collect actually ships and
```
swashtek, line 11: What are the complete set of supported and shipped targets? Should the configuration include Linux arm64, Windows arm64, or any macOS targets?

.github/workflows/cargo-deny.yml
```diff
@@ -0,0 +1,31 @@
...
+    # Installed from source to avoid taking on a third-party GitHub Action;
+    # `--locked` keeps the build reproducible.
+    - name: Install cargo-deny
+      run: cargo install cargo-deny --locked
+
+    # There is no workspace-level Cargo.toml, so each crate root is checked
+    # against the shared repo-root deny.toml.
+    - name: 'one_collect: cargo deny check'
+      run: cargo deny --manifest-path one_collect/Cargo.toml --config deny.toml check
+
```
swashtek, line 37: Should this be a required check for every PR, including PRs that do not modify dependency manifests or lockfiles?

Because `cargo deny check` evaluates the entire resolved dependency graph and could block an unrelated code change. 
If full-graph enforcement is intentional, could we document that policy?

deny.toml
utpilla: Updated.

deny.toml
brianrob: We currently build:

x86_64-pc-windows-msvc
x86_64-unknown-linux-gnu
aarch64-unknown-linux-gnu
x86_64-unknown-linux-musl
aarch64-unknown-linux-musl

deny.toml
utpilla: Done. Our internal crates link by path without a version (e.g. `engine = { path = "engine" }`), which cargo-deny treats as a `*` wildcard. `allow-wildcard-paths = true` exempts those local links while registry wildcards still fail; it only applies to non-published crates, so I marked the internal `engine`/`record-trace`/ffi crates `publish = false`.

utpilla: Done. Our internal crates link by path without a version (e.g. `engine = { path = "engine" }`), which cargo-deny treats as a `*` wildcard. `allow-wildcard-paths = true` exempts those local links while registry wildcards still fail; it only applies to non-published crates, so I marked the internal `engine`/`record-trace`/`ffi` crates `publish = false`.

deny.toml
utpilla: Updated.

deny.toml
utpilla: > I'd only want this sort of thing to run if it were limited to dependencies added in the current PR.

I want to make sure we don't lose the most useful part of advisory scanning: catching newly-disclosed advisories against dependencies we already have (they're often found long after a dep is added). Limiting the scan to only deps added in a PR would give that up.

It may not be a big deal to just keep advisories blocking on PRs too. If an unrelated older dep gets an advisory, we could send a separate PR to fix it first before merging open PRs. Or if we feel that it's not crucial, the maintainers will have admin override to merge past a failing check. The advisory can be addressed post merge of the open PRs.

What do you think?

.github/workflows/cargo-deny.yml
utpilla: Similar comment, replied here: https://github.com/microsoft/one-collect/pull/323#discussion_r3787573595

deny.toml
brianrob: 100% makes sense that we want to keep looking for new advisories in-perpetuity.  I think the thing I don't want is to have a change that doesn't touch dependencies fail because of this - In my opinion, it just adds confusion.  I wonder if there is a compromise where we run this regularly against main and get issues filed or notifications when there is an issue.

Would also like to get feedback from @beaubelgrave on this one as well.

utpilla: There is a middle option. We could keep the full-graph scan, but split where it blocks:

- `bans` / `licenses` / `sources` stay as required PR checks: these only change when someone edits dependencies, so they're deterministic and won't spontaneously break an unrelated PR.
- `advisories` moves to a scheduled workflow (e.g. daily + on push to `main`) that still scans the entire graph. A newly published advisory fails that scheduled run (and we can have it open/update a tracking issue) instead of turning a random contributor's PR red.

This keeps full awareness of new advisories on old deps while unblocking day-to-day code motion. Concretely it's a small change to `cargo-deny.yml`: the PR job runs `cargo deny check bans licenses sources`; a new scheduled job runs `cargo deny check advisories`.

brianrob: This option sounds good to me.

beaubelgrave: I like this option as well, let's go with that middle option.

utpilla: Implemented as agreed.

.github/workflows/cargo-deny-advisories.yml
```diff
@@ -0,0 +1,37 @@
...
+
+    steps:
+    - uses: actions/checkout@v3
+
+    - name: Install cargo-deny
+      run: cargo install cargo-deny --locked
+
+    # There is no workspace-level Cargo.toml, so each crate root is checked
+    # against the shared repo-root deny.toml.
+    - name: 'one_collect: cargo deny check advisories'
```
brianrob: We should make sure that both of these run.  Right now, if the first one fails, the second one won't run.

.github/workflows/cargo-deny-advisories.yml
```diff
@@ -0,0 +1,37 @@
...
+      run: cargo deny --manifest-path one_collect/Cargo.toml --config deny.toml check advisories
+
+    - name: 'record-trace: cargo deny check advisories'
```
brianrob: If either of these results in a failure, we should file an issue and auto-assign it to ourselves (either directly or via a group) to make sure that we get an e-mail and have a tracking issue.

brianrob (commented):
Overall, looks good.  A couple of comments.

.github/workflows/cargo-deny-advisories.yml
utpilla: Fixed. Both crate roots now run in a single step that aggregates the exit status, so record-trace is always scanned even when one_collect reports an advisory, and the step still fails if either does.

.github/workflows/cargo-deny-advisories.yml
utpilla: Added. On failure the workflow files a tracking issue (label `cargo-deny-advisories`) assigned to @brianrob and @beaubelgrave, with the cargo-deny output attached, so you get an email and a tracking issue. To avoid a new issue every day it comments on the existing open one instead of opening duplicates, and a later passing run auto-closes it.

brianrob (commented):
Looking good.  A couple of small changes and we should be ready to merge.

.github/workflows/cargo-deny-advisories.yml
```diff
@@ -0,0 +1,87 @@
...
+jobs:
+  advisories:
+    name: Cargo Deny Advisories
+    runs-on: ubuntu-latest
+    env:
+      GH_TOKEN: ${{ github.token }}
+      ISSUE_LABEL: cargo-deny-advisories
+
+    steps:
+    - uses: actions/checkout@v3
```
brianrob: Please update this to use the full hash. See https://github.com/microsoft/one-collect/commit/bc9aef7f48497446ab895776612e13976861a06d for details.

.github/workflows/cargo-deny.yml
```diff
@@ -0,0 +1,39 @@
...
+env:
+  CARGO_TERM_COLOR: always
+
+jobs:
+  cargo-deny:
+    name: Cargo Deny
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v3
```
brianrob: Same here.

.github/workflows/cargo-deny-advisories.yml
utpilla:
```suggestion
    - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
```

.github/workflows/cargo-deny.yml
utpilla:
```suggestion
    - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
```

.github/workflows/cargo-deny-advisories.yml
utpilla: Fixed.

.github/workflows/cargo-deny.yml
utpilla: Fixed.

brianrob (approved):
LGTM.  Thanks!

PR #326, opened 2026-08-14
Refactor SchemaCache methods to unify schema indexing and insertion logic

swashtek:
## Summary

This is a follow-up to [PR #313](https://github.com/microsoft/one-collect/pull/313), created from the review discussion about simplifying the `SchemaCache` API and centralizing schema caching behavior.

The change unifies schema lookup and insertion for TraceLogging and manifest schemas while preserving their source-specific key maps and existing negative-cache behavior.

## Motivation

During review of PR #313, it was suggested that the cache preserve generic `index_of` and `insert` methods rather than exposing separate TraceLogging- and manifest-specific methods.

The follow-up discussion also identified that cache lookup and insertion should be centralized before dispatching to `decode_tracelogging` or `decode_manifest`. This keeps the decode methods focused on performing the actual decode and leaves cache selection in one place.

See the originating review discussion:
- [Centralize caching behavior](https://github.com/microsoft/one-collect/pull/313#discussion_r3641948556)
- [Unify `index_of` and `insert`](https://github.com/microsoft/one-collect/pull/313#discussion_r3641975944)

## Changes

- Add a `SchemaSource`-based `SchemaCache::index_of` method.
- Add a unified `SchemaCache::insert` method for TraceLogging and manifest schemas.
- Move source-specific cache selection into `SchemaCache`.
- Centralize positive-cache lookup and insertion in the decoder flow.
- Preserve separate TraceLogging and manifest maps because their key types differ.
- Preserve separate 32-bit and 64-bit cache key spaces.
- Keep manifest negative-cache handling separate because TraceLogging has no equivalent permanently unsupported state.
- Ensure duplicate inserts return the existing index without growing the schema arena.
- Update unit tests for the unified API, duplicate insertion behavior, source isolation, and pointer-width separation.

## Scope

This is an internal API-shape refactor with no intended behavioral change. It is separated from PR #313 so the manifest and caching functionality can remain independently reviewable.

### DIFF
--- a/one_collect/src/etw/tdh.rs
+++ b/one_collect/src/etw/tdh.rs
@@ -275,17 +275,27 @@ impl SchemaCache {
         }
     }
 
-    /// Returns the index of a cached TraceLogging schema, or `None` on a
-    /// miss.  The key is hashed exactly once.
-    fn index_of_tl(&self, key: &[u8], is_32bit: bool) -> Option<usize> {
-        let map = if is_32bit { &self.tl_32 } else { &self.tl_64 };
-        map.get(key).copied()
-    }
-
-    /// Returns the index of a cached manifest schema, or `None` on a miss.
-    fn index_of_manifest(&self, key: &ManifestEventKey, is_32bit: bool) -> Option<usize> {
-        let map = if is_32bit { &self.manifest_32 } else { &self.manifest_64 };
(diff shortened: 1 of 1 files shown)


utpilla (approved):

brianrob (approved):
LGTM.  Thanks!

Issue #327, opened 2026-08-18
Record-trace hangs for several minutes


srveniga:
When using a low memory limit with record-trace as below, the command hangs for around 20 mins before exiting as shown below.

```
# date && ./latest_record_trace_bin --script-file ./aug_record_trace/record-trace/net-xmit.script --max-memory 1 --live && date
Fri Aug 14 20:24:05 UTC 2026
Memory limit reached.
Recording started.  Press CTRL+C to stop.
 
Recording stopped.
Resolving symbols.
Finished recording trace.
Trace written to /tmp
Fri Aug 14 20:48:32 UTC 2026
```
trace.log file shows below warnings, implying that the writes for the capture environment thread are timing out, waiting for space. Not hitting this on my dev box but hitting it consistently on a production system.

```
2026-08-14T22:23:26.807015Z  WARN one_collect::perf_event::rb: InProcessRingBufWriter::write: timed out waiting for space, dropping record_len=120
2026-08-14T22:23:26.807051Z  WARN one_collect::perf_event::rb: InProcessRingBufWriter::write: insufficient space for record_len=96, waiting
2026-08-14T22:23:26.907738Z  WARN one_collect::perf_event::rb: InProcessRingBufWriter::write: timed out waiting for space, dropping record_len=96
2026-08-14T22:23:26.907775Z  WARN one_collect::perf_event::rb: InProcessRingBufWriter::write: insufficient space for record_len=96, waiting
2026-08-14T22:23:27.008420Z  WARN one_collect::perf_event::rb: InProcessRingBufWriter::write: timed out waiting for space, dropping record_len=96
2026-08-14T22:23:27.008448Z  WARN one_collect::perf_event::rb: InProcessRingBufWriter::write: insufficient space for record_len=96, waiting
```

The binary built with the change in srveniga:srveniga/fix_memlimit_hang fixes this and I can see the command exiting within a second on the same production system, as expected.

```
# date && ./fixed-record-trace --script-file ./aug_record_trace/record-trace/net-xmit.script --max-memory 1 --live && date
Fri Aug 14 20:51:30 UTC 2026
Memory limit reached.
Recording started.  Press CTRL+C to stop.
 
Recording stopped.
Resolving symbols.
Finished recording trace.
Trace written to /tmp
Fri Aug 14 20:51:31 UTC 2026
```

PR #328, opened 2026-08-18
Fix for issue #327

srveniga:
Continue parsing the COMM, MMAP events from the environment capture thread so that there is enough space in ring buffer to add all the events from the system. This snippet avoids a several minute hang on systems with high number of COMM/MMAP events. More details in Issue #327


beaubelgrave (approved):
LGTM

PR #332, opened 2026-09-02
Add total event buffer size policy

hoyosjs:
Add `--buffersize <MB>` to `record-trace` and expose `with_buffer_size_bytes` for callers that specify a total event buffer data capacity.

Normalize the requested capacity according to each platform's underlying buffer model. Linux accounts for all online CPU rings and the in-process ring, then rounds down to a valid power-of-two page count. Windows uses the ETW buffer pool size and constrains individual buffers to ETW's supported range. The computed buffer count and per-buffer size are logged so the resulting capacity is observable.

Keep `with_per_cpu_buffer_bytes` as the direct per-buffer upper-bound API. The two sizing models are mutually exclusive, so a script that uses both is rejected rather than silently discarding one of the requested sizes. Repeating either one overrides the earlier value, and a script setting overrides `--buffersize`. Zero and negative sizes are rejected.

The requested total replaces the default capacity rather than raising it. Platform minimums and normalization can cause the resulting capacity to differ from the request.

### Breaking changes

- `with_per_cpu_buffer_bytes` is consistently treated as an upper bound, so Linux values that are not power-of-two page counts now round down rather than up.
- `EtwSession::with_per_cpu_buffer_bytes` clamps to the ETW-supported range of 4 KB to 16 MB. A size above that range previously failed session startup with `StartTraceW` error 87 and now uses 16 MB.

No existing public Rust signatures are removed or changed.

### Validation

Linux buffer sizing was reproduced with `record-trace` at multiple capacities and the resulting buffer count and per-buffer size were confirmed from logs. The Windows and ETW paths were reviewed at the source level because a Windows target was not available in the validation environment.

Related: dotnet/diagnostics#5995


record-trace/engine/src/recorder.rs
```diff
@@ -310,6 +310,14 @@ impl Recorder {
                 debug!("Using default configuration");
                 UniversalExporter::new(settings)
             }
+        };
+
+        let universal = match self.args.per_cpu_buffer_size_bytes() {
+            Some(bytes) => {
+                info!("Configuring per-CPU event buffer: bytes={}", bytes);
+                universal.with_per_cpu_buffer_bytes(bytes)
```
brianrob: This brings up an interesting question.  I don't think this API actually does what we expect.  For perf\_events it converts to pages per CPU, which is correct (or at least close).  For ETW, this value is used as the buffer size, and then there is a min and max buffer count based on the number of CPUs.  So while this change seems very reasonable, I think we might need to address this issue at the API level before we expose it via record-trace.  I'm inclined to have a high watermark buffer size for the entire session and have the Universal layer choose reasonable values at the perf\_events/ETW layer that respect it.  For example:

perf\_events: Take the high watermark, divide it by the number of processors and use that as the per-processor buffer size.

ETW: Take the high watermark and divide it by the value of MaxBuffers.

We can also ensure buffers are of a minimum size as well, so that scripts will more reasonably scale to higher core count machines.  Users of the perf\_events or ETW layer directly will have complete control over how they configure the buffers, but I think we'll need to set some kind of policy for Universal.  @beaubelgrave, what do you think?

brianrob (commented):
Thanks @hoyosjs.  You may have found a hidden landmine in the code here.  :)

record-trace/engine/src/recorder.rs
beaubelgrave: Yeah, that sounds good at the Export / Record-Trace level, folks going deeper into the stack can choose how to optimize. As long as we have something easily explainable and has the same intentions on all OS parts, I'm good with it.

one_collect/src/helpers/exporting/scripting.rs
```diff
@@ -917,6 +917,16 @@ impl ScriptedUniversalExporter {
...
 
+        let fn_exporter = self.export_swapper();
+        self.rhai_engine().register_fn(
+            "with_buffer_size_bytes",
+            move |size: i64| {
+                info!("Setting total buffer size: bytes={}", size);
+                fn_exporter.borrow_mut().swap(|exporter| {
+                    exporter.with_buffer_size_bytes(size as usize)
+                });
+            });
```
swashtek, line 928: The buffer size in bytes can accept negative values and casts as `usize`. `with_buffer_size_bytes(-1)` becomes very large unsigned value. Can we add check for non-positive values, or return error ?

```
            move |size: i64| {
            move |size: i64| {
                if size <= 0 {
                    return Err("buffer size must be positive".into());
                }
                 info!("Setting total buffer size: bytes={}", size);
                 fn_exporter.borrow_mut().swap(|exporter| {
                     exporter.with_buffer_size_bytes(size as usize)
                 });
             });

```

one_collect/src/helpers/exporting/os/linux.rs
```diff
@@ -1777,7 +1777,11 @@ impl UniversalExporterOSHooks for UniversalExporter {
 
         let page_size = unsafe { libc::sysconf(libc::_SC_PAGE_SIZE) as usize };
 
-        let page_count = self.cpu_buf_bytes() / page_size;
+        let cpu_count = match &settings.target_cpus {
+            Some(cpus) => cpus.iter().copied().collect::<std::collections::HashSet<_>>().len(),
+            None => ExportMachine::cpu_count() as usize,
+        };
+        let page_count = self.buffer_bytes(cpu_count) / page_size;
```
swashtek, line 1784: The calculation does not preserve the requested total buffer size on linux because perf rounds the page count up to next power of two in `ring_data_bytes()`. 
For example, for a 1 MiB request across 3 CPUs, this path passes 85 pages per CPU, while `ring_data_bytes()` allocates 128 pages per CPU (~ 1.5 MiB). 
It would be good to consider normalizing it here against perf rounds rules or clarify API contract and coverage to this.

one_collect/src/helpers/exporting/scripting.rs
hoyosjs: Yeah - sorry. I am cleaning this up a bit with the new suggested change. Didnd't mean to push yet.

brianrob (commented):
Thanks @hoyosjs.  This is looking good.  As I reviewed it, I ran into another small structural thing that we might want to change.

@BeauBelgrave, can you please see my comment below regarding moving the API to `ExportSettings`.

one_collect/src/etw/mod.rs
```diff
@@ -495,10 +501,17 @@ impl EtwSession {
...
+    /// ETW fails to start a session whose buffer size falls outside its
+    /// supported range, so the requested size is clamped instead of being
+    /// passed through. See `EVENT_TRACE_PROPERTIES`:
+    /// <https://learn.microsoft.com/en-us/windows/win32/api/evntrace/ns-evntrace-event_trace_properties>
     pub fn with_per_cpu_buffer_bytes(
         mut self,
         bytes: usize) -> Self {
-        self.cpu_buf_kb = (bytes / 1024) as u32;
+        self.cpu_buf_kb = (bytes / 1024)
+            .clamp(MIN_BUFFER_SIZE_KB, MAX_BUFFER_SIZE_KB) as u32;
```
brianrob: Can you please add logging here so that if the value is modified that the user can look at a log to know this?

one_collect/src/helpers/exporting/os/linux.rs
```diff
@@ -38,6 +38,16 @@ pub type Session = PerfSession;
 /* OS Specific Session Builder Type */
 pub type SessionBuilder = RingBufSessionBuilder;
 
+fn underlying_perf_buffer_count() -> usize {
+    (crate::perf_event::rb::cpu_count() as usize).saturating_add(1)
+}
+
+fn perf_buffer_page_count(buffer_bytes: usize) -> usize {
```
brianrob: Please add a quick comment here to explain this calculation (especially the last line).

one_collect/src/helpers/exporting/scripting.rs
```diff
@@ -8,14 +8,33 @@ use crate::event::*;
 use rhai::{CustomType, TypeBuilder, Engine, EvalAltResult};
 use tracing::{info, error};
 
+#[derive(Clone, Copy, PartialEq)]
+enum ScriptBufferSize {
```
brianrob: Can we just use the `BufferSize` enum directly here instead of having a second one?  It looks like `ScriptBufferSize` isn't actually exposed to the script writer.

one_collect/src/helpers/exporting/scripting.rs
```diff
@@ -50,6 +69,38 @@ impl UniversalExporterSwapper {
         }
     }
 
+    /// Applies one of the two mutually exclusive buffer sizing models. Mixing
+    /// them is rejected rather than letting one of the requested sizes be
+    /// silently discarded, which would leave the trace running with a capacity
+    /// the script never asked for. Repeating the same model overrides the
+    /// previously requested size.
+    fn set_buffer_size(
```
brianrob: Thinking about this scenario a bit more as I review the code - I am wondering if we should move the buffer configuration into `ExportSettings` and then remove the existing `with_per_cpu_buffer_bytes`.  As it stands now, you can set the value multiple times, but this doesn't change the configuration during tracing, it just changes the configuration while it's being built-up before the session starts.  If that's the case, it seems like we should just have this be part of the `ExportSettings`.  @BeauBelgrave, what are your thoughts on this?  I realize this would create a breaking change if we remove the existing API, but it seems like adding a new one is somewhat complicated, and the policy around when/how often it can be called can all go away if we put it in `ExportSettings`.

